PR #12, opened 2020-10-16
Add MoveSwapBlock and ResetStateCrystal

coloursofnoise:
Pretty sure these are both feature-complete and I just forgot to merge them.

Adds a MoveSwapBlock from @Viv-0 that I cleaned up a bit.
Adds a ResetStateCrystal also from Viv that resets the player state to StNormal when collected


coloursofnoise:
Also moves the custom spritebank to `Graphics/CommunalHelper/Sprites.xml`

PR #18, opened 2021-01-07
added grouped release of timed trigger spikes

Viv-0:


src/Entities/Misc/TimedTriggerSpikes.cs
```diff
@@ -108,27 +107,49 @@ private struct SpikeInfo {
...
+        protected bool Grouped {
+            get {
+                return grouped && CommunalHelperModule.maxHelpingHandLoaded;
+            }
+        }
+        protected bool Triggered = false;
+
         public static Entity LoadUp(Level level, LevelData levelData, Vector2 offset, EntityData entityData) {
+            if (entityData.Bool("Grouped", false) && !CommunalHelperModule.maxHelpingHandLoaded) {
+                throw new Exception("Grouped Timed Trigger Spikes attempted to load without Max's Helping Hand as a dependency.");
```
coloursofnoise, line 122: Looking at it, probably move this into the EntityData constructor if that still gets caught by the LoadingThread try/catch

src/CommunalHelperModule.cs
```diff
@@ -21,6 +21,8 @@ public class CommunalHelperModule : EverestModule {
 
         public static SpriteBank SpriteBank => Instance._SpriteBank;
         public SpriteBank _SpriteBank;
+
+        public static bool maxHelpingHandLoaded { get; private set; }
```
coloursofnoise, line 25: Make this `internal` and rename it to `MaxHelpingHandLoaded`

coloursofnoise (changes requested):
Looks good other than a couple nitpicky things

coloursofnoise (approved):

PR #22, opened 2021-02-04
Grouped trigger spikes

vivianlonging:
Rainbow Grouped Trigger Spike


src/Entities/Misc/TimedTriggerSpikes.cs
```diff
@@ -312,6 +345,11 @@ public TimedTriggerSpikes(Vector2 position, int size, Directions direction, stri
         }
 
         public override void Update() {
+            if (rainbow) {
+                if (crystalSpinner == null) { crystalSpinner = new CrystalStaticSpinner(Vector2.One * -1000 + offset, false, CrystalColor.Rainbow); }
+                if (Scene == null) { return; }
+                if (!Scene.Tracker.GetEntities<CrystalStaticSpinner>().Contains(crystalSpinner)) { Scene.Add(crystalSpinner); }
+            }
```
coloursofnoise, line 352: This could definitely be improved. 
Under what circumstances would `crystalSpinner` be `null` at this point? the same for `Scene`, since the entity should then no longer be updated.
If you want to check if `crystalSpinner` is part of the current scene it's probably best to do `crystalSpinner.Scene == Scene`.

Ahorn/entities/TimedTriggerSpikes.jl
```diff
@@ -1,11 +1,10 @@
...
-
+@mapdef Entity "CommunalHelper/TimedTriggerSpikesUp" TimedTriggerSpikesUp(x::Integer, y::Integer, width::Integer=Maple.defaultSpikeWidth, type::String="default", Delay::Number=0.4, WaitForPlayer::Bool=false, Grouped::Bool=false, Rainbow::Bool=false)
+@mapdef Entity "CommunalHelper/TimedTriggerSpikesDown" TimedTriggerSpikesDown(x::Integer, y::Integer, width::Integer=Maple.defaultSpikeWidth, type::String="default", Delay::Number=0.4, WaitForPlayer::Bool=false, Grouped::Bool=false, Rainbow::Bool=false)
+@mapdef Entity "CommunalHelper/TimedTriggerSpikesLeft" TimedTriggerSpikesLeft(x::Integer, y::Integer, height::Integer=Maple.defaultSpikeHeight, type::String="default", Delay::Number=0.4, WaitForPlayer::Bool=false, Grouped::Bool=false, Rainbow::Bool=false)
+@mapdef Entity "CommunalHelper/TimedTriggerSpikesRight" TimedTriggerSpikesRight(x::Integer, y::Integer, height::Integer=Maple.defaultSpikeHeight, type::String="default", Delay::Number=0.4, WaitForPlayer::Bool=false, Grouped::Bool=false, Rainbow::Bool=false)
```
coloursofnoise, line 7: Please update the en_gb.lang file to add a "Rainbow" attribute tooltip

coloursofnoise (changes requested):
In addition to these, please run a Code Cleanup to get rid of unnecessary `base.` and casts, and use the static `Depths` and `Sfx` classes for your depth/sfx constants

coloursofnoise (approved):

PR #31, opened 2021-03-17
Manual Cassette Controller

coloursofnoise:
Adds a controller entity that can be added to a room to allow the player to manually cycle the cassette blocks.

Attempts to remove/prevent the addition of a `CassetteBlockManager`, and throws an exception if one is still present.

Features a "secret" in the form of a 1/100 chance to use a different icon in Ahorn


catapillie:
Looks good.

PR #32, opened 2021-04-05
Add Input Flag Controller

coloursofnoise:
A controller entity that allows you to specify a list of flags that can be grouped into a sequence.
Upon pressing the controller button it will cycle through the flags and either enable them or toggle them based on behaviour specified in Ahorn.

Also restructures Controllers to use a common parent class to cut down on reused code and hooks


catapillie:
looks good 👍
missing the description in the game banana page but it's fine for now

PR #36, opened 2021-04-25
Add SoundSurfaceIndex Panel

coloursofnoise:
A panel that can be attached to a solid to change the `SoundSurfaceIndex` used by it. 
Currently does not work with SolidTiles or other solids that set `AllowStaticMovers = false` because you can just change the index used by a particular tileset instead.

Also abstracts out some of the "panel" behaviour from `DreamTunnelEntry` so that it can be reused


catapillie:
Removed the Down placement, doesn't seem necessary

PR #45, opened 2021-05-22
Fix CassetteZipMover liftboost when deactivating

coloursofnoise:
Also cleans up CassetteZipMover code (three different fields for tracking the nodes?) and improves the "spark" particles for nodes

Should keep full functionality (including `dynamicHitbox`) for other CassetteBlocks


catapillie (approved):
Looks good, thanks for cleaning up some stuff as well

PR #64, opened 2021-07-01
Fix MoveSwapBlock LiftSpeed issues

### LINKED ISSUE #62: MoveSwapBlocks don't give upwards LiftBoost when moving
Using a MoveSwapBlock that moves horizontally and swaps vertically, vertical liftboost works properly when the block is stationary, but not when it is moving.

https://user-images.githubusercontent.com/53288101/122601041-8e9db680-d025-11eb-8368-f25cf4f79d40.mp4

coloursofnoise:
- Closes #62 (both issues)


catapillie (approved):
Looks all good.

Issue #21, opened 2021-01-29
Dream Jelly Fish


catapillie:
![](https://cdn.discordapp.com/attachments/740715134766415912/804544486599950356/DreamJelly.gif)
Self-explanatory, could be useful for people that struggle with performing jellyvators

> And I think if this is developed there's a benefit to making it be an accessibility option to people who can't jellyvate by making a gameplay option to replace all jellyfish in a room with space jellyfish that don't refill dashes and don't convert, meaning you just need to dash up and hold grab instead of jellyvating

> Would suggest making it so that space jellyfish do give back dashes, but once you dash through a space jellyfish it becomes "inactive" meaning you can't dash through it; but once the jellyfish touches certain things (ground, springs, maybe it'd be interesting if it touched a space jam too?) it "reactivates" and allows you to dash through it again

Lots of ideas to consider, could be fun

PR #65, opened 2021-07-17
Add Dream Jellyfish

catapillie:
Implements #21.

Dream Jellyfishes use their own renderer that itself uses its own `VirtualRenderTarget`, looping through `DreamJellyfish` entities in the level and rendering them on the render target, then being stamped onto the Gameplay render target, after which it'll clear itself, and move on to the next entity.

This implementation also includes a new component, `DreamDashCollider`, which takes in a `Collider` and an eventual `onExit` delegate (`Action<Player>`). Once added to an entity, the component will allow the player to dream dash through the collider, relative to the entity's position. If the player exits the dream dash state, and the latest object it went through was a `DreamDashCollider`, its `onExit` delegate is invoked with the player object.


catapillie:
don't worry about the audio bank conflicts, i'll rebuild and push on `dev`.. / just pick the latest one

src/Entities/DreamStuff/DreamDashCollider.cs
```diff
@@ -0,0 +1,116 @@
...
+                    playerData[Player_canEnterDreamDashCollider] = false;
+                    collider.OnExit?.Invoke(self);
+                }
+            }
+            orig(self);
+        }
+
+        private static void Player_DreamDashUpdate(ILContext il) {
+            ILCursor cursor = new ILCursor(il);
+            if (cursor.TryGotoNext(MoveType.Before, instr => instr.MatchStloc(1))) {
```
coloursofnoise: I'd like to match against something a little more reliable than just a `stloc`, is there a method or field or something near it that could be used as reference?

src/Entities/DreamStuff/DreamDashCollider.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// Checks if the player is colliding with this component.
+        /// </summary>
+        /// <param name="player">The player instance.</param>
+        private bool Check(Player player) {
+            if (Active && Entity != null &&
+                player.GetData().Data.TryGetValue(Player_canEnterDreamDashCollider, out object canEnter) && canEnter.Equals(true)) {
+                
+                Collider collider = Entity.Collider;
+
+                Entity.Collider = Collider;
```
coloursofnoise, line 42: this should only be done if `Collider` is not `null`

src/Entities/DreamStuff/DreamDashCollider.cs
```diff
@@ -0,0 +1,116 @@
...
+    class DreamDashCollider : Component {
+        public static readonly Color ActiveColor = Color.Teal;
+        public static readonly Color InactiveColor = Calc.HexToColor("044f63"); // darker teal
+
+        public Collider Collider;
+        public DreamBlockDummy Dummy;
+
+        public Action<Player> OnExit;
+
+        public DreamDashCollider(Collider collider, Action<Player> onExit_player_dreamJumped = null)
```
coloursofnoise: `onExit_player_dreamJumped` is a very long parameter name that seems to imply it will only be invoked if the player dreamjumps? maybe shorten the name and explain it in an xmldoc comment instead

src/Entities/DreamStuff/DreamJellyfish.cs
```diff
@@ -0,0 +1,268 @@
...
+
+        private readonly DreamDashCollider dreamDashCollider;
+        public bool AllowDreamDash {
+            get {
+                return dreamDashCollider.Active;
+            }
+            set {
+                dreamDashCollider.Active = value;
+            }
+        }
```
coloursofnoise:
```suggestion
        public bool AllowDreamDash {
            get => dreamDashCollider.Active;
            set => dreamDashCollider.Active = value;
        }
```

src/Entities/DreamStuff/DreamJellyfish.cs
```diff
@@ -0,0 +1,268 @@
...
+                Particles[i].TimeOffset = Calc.Random.NextFloat();
+
+                Particles[i].DisabledColor = Color.LightGray * (0.5f + Particles[i].Layer / 2f * 0.5f);
+                Particles[i].DisabledColor.A = 255;
+
+                Particles[i].EnabledColor = Particles[i].Layer switch {
+                    2 => Calc.Random.Choose(Calc.HexToColor("5b6ee1"), Calc.HexToColor("CC3B3B"), Calc.HexToColor("7daa64")),
+                    1 => Calc.Random.Choose(Calc.HexToColor("5fcde4"), Calc.HexToColor("7fb25e"), Calc.HexToColor("E0564C")),
+                    _ => Calc.Random.Choose(Calc.HexToColor("FFEF11"), Calc.HexToColor("FF00D0"), Calc.HexToColor("08a310")),
+                };
```
coloursofnoise: you separated out the `CustomDreamBlock.DreamColors` but then didn't make use of them here

coloursofnoise (changes requested):
I don't really have the ability to comment on the `DreamJellyfishRenderer`, but other than a couple nitpicks and a wish for a little more documentation/comments the PR looks good :+1:

src/Entities/DreamStuff/DreamDashCollider.cs
catapillie: I was going to make the delegate require a bool that is `true` when the player dreamjumps, but then I found out that you could dreamjump after it was called, so I removed it

src/Entities/DreamStuff/DreamDashCollider.cs
catapillie: are you sure this is needed? if the entity doesn't have a collider, then there is no collision. i'll just add the check so that it doesn't bother doing it anyway
edit: nevermind i misunderstood what you meant

coloursofnoise (approved):

PR #67, opened 2021-07-23
Add chained block entities and decorative chains

### LINKED ISSUE #44: Chained Falling Block
A FallingBlock that will fall a specified distance before being stopped by chains.
## Ideas
- Realistic dangle around after falling, like how objects would do in real life

- Indicator for height, togglable if doesn't look great

- Chain breaks 
  - After standing on the block for too long when it's held in air
  - Immediately or with short delay just after first impact
  - When the player dashes into it (or wavedashes, hypers off of it)
  - Two fall sequences with different height values

- Block breaks if impact against solid

example image

catapillie:
Closes #44.

And adds:
 * `Chain` (decoration entity).
 * `ChainedKevin`.
 * `chained` variant of `DreamFallingBlock`.


src/CommunalHelperModule.cs
```diff
@@ -265,6 +271,18 @@ public static class Util {
...
+        }
+
+        public static Vector2 MoveBlockDirectionToVector(MoveBlock.Directions direction) {
+            return direction switch {
+                MoveBlock.Directions.Up => -Vector2.UnitY,
+                MoveBlock.Directions.Down => Vector2.UnitY,
+                MoveBlock.Directions.Left => -Vector2.UnitX,
+                _ => Vector2.UnitX,
+            };
+        }
```
coloursofnoise: This is already implemented as an extension method:
https://github.com/CommunalHelper/CommunalHelper/blob/78660e47cd9d247e6e9527e6a7e24c2a892289ab/src/Extensions.cs#L258-L273

src/Entities/ChainStuff/ChainedKevin.cs
```diff
@@ -0,0 +1,255 @@
+﻿using Celeste.Mod.Entities;
+using Microsoft.Xna.Framework;
+using Mono.Cecil.Cil;
+using Monocle;
+using MonoMod.Cil;
+using MonoMod.Utils;
+using System;
+using System.Collections.Generic;
+using static Celeste.MoveBlock;
```
coloursofnoise: To only import a type as a specific name, which is less likely to produce name conflicts, use
```suggestion
using Directions = Celeste.MoveBlock.Directions;
```

src/Entities/ChainStuff/ChainedKevin.cs
```diff
@@ -0,0 +1,255 @@
...
+                chainedKevin.vectorDirection = Util.MoveBlockDirectionToVector(chainedKevin.direction = data.Enum("direction", Directions.Right));
+                chainedKevin.centeredChain = data.Bool("centeredChain");
+            }
+            orig(self, data, offset);
+        }
+
+        private static void IL_CrushBlock_ctor_Vector2_float_float_Axes_bool(ILContext il) {
+            ILCursor cursor = new ILCursor(il);
+            if (cursor.TryGotoNext(instr => instr.MatchLdstr("objects/crushblock/block")) &&
+                cursor.TryGotoNext(MoveType.After, instr => instr.MatchStloc(0))) {
```
coloursofnoise: does it matter exactly where this code is run, as long as it's after the lists are initialized? if not then you could just match against the string (this could even just be done as part of the `AddImage` hook by checking if the `chainedKevin` fields are null, and initializing them if they are)

src/Entities/DreamStuff/DreamFallingBlock.cs
```diff
@@ -15,11 +35,39 @@ public class DreamFallingBlock : CustomDreamBlock {
...
+        private bool indicator, indicatorAtStart;
+        private float pathLerp;
+
+        private SoundSource rattle;
+
         public DreamFallingBlock(EntityData data, Vector2 offset)
             : base(data.Position + offset, data.Width, data.Height, data.Bool("featherMode"), data.Bool("oneUse"), GetRefillCount(data), data.Bool("below")) {
             noCollide = data.Bool("noCollide", false);
 
+            chained = data.Bool("chained");
```
coloursofnoise, line 55: would it be feasible to implement this as a subclass of `DreamFallingBlock` instead?

coloursofnoise (changes requested):
If it's something that's decided on, refactoring the ChainedDreamFallingBlock can be done after the update

src/Entities/DreamStuff/DreamFallingBlock.cs
catapillie: I'd like to do this for 1.9.1, we don't have time anymore

catapillie: This will be worked on for #69.

PR #78, opened 2021-09-29
More Panels

coloursofnoise:
Add three new panels:
- FrictionlessPanel: Prevents the player from being dragged by moving solids when standing or climbing on them
- DashCollision Panel: Allows setting the dash collision result for a surface
- Bouncy Panel: ExtendedVariants Bouncy Everywhere mode, but not Everywhere.


catapillie:
- Small detail: panels with `overrideAllowStaticMovers = false` that are placed on solids that don't allow static movers remain active for one frame, and when you reload the map next to one, you might trigger its behavior (for instance, the bouncy panel bounces you off and then disappears).

- Also, I can't get them to work on `SolidTiles`, whether I tick `overrideAllowStaticMovers` or not... Do you know about that?

- The DashCollision Panel doesn't work very well when placed on floaty blocks/solids.

- No "down" orientation placement for the Bounce Panel? Dashing from the bottom, up or diagonally, and then bouncing back would be a pretty neat movement in my opinion 

- this (fixed by rescaling the panel though)

https://user-images.githubusercontent.com/59423464/138496183-d2c1e8e0-53c1-4de1-b2e6-e39771b7628c.mp4

coloursofnoise:
> panels with `overrideAllowStaticMovers = false` that are placed on solids that don't allow static movers remain active for one frame

not sure that it's worth fixing this, since the I think the alternative would be that panels would be disabled for the first active frame.

> Also, I can't get them to work on `SolidTiles`, whether I tick `overrideAllowStaticMovers` or not... Do you know about that?

Fixed a possibly related bug where the platform's `OnDashCollide` delegate wasn't getting overwritten properly

> The DashCollision Panel doesn't work very well when placed on floaty blocks/solids.

`FloatySpaceBlock` assigns to `OnDashCollide` during `Awake`, so this is probably load-order dependent. Unsure whether it's worth fixing for now.

> No "down" orientation placement for the Bounce Panel? Dashing from the bottom, up or diagonally, and then bouncing back would be a pretty neat movement in my opinion

Celeste does not have a built in Bounce method for a downwards bounce, and one is not currently provided by ExtendedVariants, which this is derived from.  
I would recommend using a DashCollisionPanel with a `Bounce` result instead (yes they're named the same, yes it is different).

> this (fixed by rescaling the panel though)

fixed, needed to account for the platform's movement when checking player collision

PR #82, opened 2021-10-15
Fix inconsistent shaking behaviour for dream falling block types

### LINKED ISSUE #81: Chained Falling Blocks Stalling
- [N] My mods and Everest install are up to date.
- [Y] I have recreated the bug with *only* CommunalHelper OR a minimum number of other mods enabled.

**Describe the bug**
Chain falling blocks have a set fall time instead of having the same fall behavior as other falling blocks. (This is also present in Viv Helper falling blocks)

**To Reproduce**
Step on and off or grab on and off a chain falling block quickly. It will stay still for longer than expected before faling.

**Expected behavior**
Normal falling blocks fall soon after they are stepped off of or let go of. This should remain the same for the chained ones.

**System Info(please complete the following information):**
 - OS: Windows

**Additional context**
Please keep this behavior as a legacy option after it is fixed. Someone has already used the bugged movement for their map.

catapillie:
Fixes and closes #81.
The addition of Chained Falling Block entities caused the `DreamFallingBlock.Triggered` field to be assigned differently.

Now, `DreamFallingBlock.HasStartedFalling` is used to know when to display the chained variant indicator, and `ChainedFallingBlock.triggered` has been renamed to `hasStartedFalling`, and isn't involved in the behavior.

### DIFF
--- a/src/Entities/ChainStuff/ChainedDreamFallingBlock.cs
+++ b/src/Entities/ChainStuff/ChainedDreamFallingBlock.cs
@@ -69,12 +69,12 @@ namespace Celeste.Mod.CommunalHelper.Entities.ChainStuff {
         public override void Update() {
             base.Update();
 
-            if (Triggered && indicator && !indicatorAtStart)
+            if (HasStartedFalling && indicator && !indicatorAtStart)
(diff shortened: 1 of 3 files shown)


coloursofnoise (approved):
Changes look reasonable to me :+1:

PR #84, opened 2021-10-18
Add extra options to Track Switch Box to allow disabling the floating behavior.

### LINKED ISSUE #83: Add floaty toggle for Track Switch Boxes
Feature Request: Add floaty/sink toggle for Track Switch Boxes
Details: a toggle for the floaty and sink code implemented in C#; asked for by Ezel in Strawberry Jam for aesthetic purposes, since this was asked for both sink and floaty, this technically includes just covering over that part of the update code within an if statement

earthwise01:
This should close #83 by adding a `canFloat`/`float` and `canBounce`/`bounce` option to Track Switch Boxes to disable the **floating** behavior and the **bouncing away from the player when dashed into** behavior respectively.

### DIFF
--- a/Ahorn/entities/TrackSwitchBox.jl
+++ b/Ahorn/entities/TrackSwitchBox.jl
@@ -6,7 +6,7 @@ using ..Ahorn, Maple
     x::Integer,
     y::Integer,
     globalSwitch::Bool=false,
-    float::Bool=true,
+    floaty::Bool=true,
     bounce::Bool=true,
 )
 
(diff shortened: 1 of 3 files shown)


catapillie:
Looks pretty good, I'll quickly test it out when I get to and merge it, should be fine.

earthwise01:
I've renamed the `float` option in the Ahorn plugin to `floaty` so that it doesn't get confused with an actual `float` and also for consistency with the Floaty Space Block.

PR #91, opened 2021-12-31
Make it so always active dream refill/dream tunnel ignores buttons, Kevins, DashBlocks, and similar objects

kfish610:
I made this behavior the default, because honestly I'm not sure why you'd want the other one? But I could pretty easily make it controlled by another setting if that's preferable. 
But basically all I did is checked if something had a OnDashCollide action, in which case dream tunnel doesn't work on it. The only exceptions were Farewell floaty blocks, since those would be kinda lame if you couldn't dash through them, and certain DashBlocks like the ones in Reflection which can only be destroyed by Kevins. Also, there's some weird behavior with buttons in the ceiling that I had to fix. Hope this looks good!


coloursofnoise:
Since this would break existing maps, it should not be the default behavior.

Also I'm a bit hesitant to hardcode the ignored entities, it might be preferable to give a way for the user to specify what should be ignored so that entities added by other mods can also be supported.

kfish610:
Oh sorry, I'm not sure why GitHub didn't notify me about this. I'll get to work adding an option for this. As for choosing which blocks get ignored, that sounds like a good idea, but I'm not really sure how I'd make that configurable? Would it be some sort of config file? Or is there a setting type that could be used.

coloursofnoise:
My thought would be to allow a comma-separated list of classnames/entity IDs for mapmakers to input for DreamRefills... but that's not very usable for the global toggle.
Maybe if it's not modified by an entity then it uses the defaults you currently have set up?

Another consideration is whether to make this a three-way toggle for
- Doesn't interact with `OnDashCollide` at all
- Respects `OnDashCollide`, does not tunnel
- Respects `OnDashCollide` *and* tunnels

The third one is possible through the two commented-out IL hooks in `DreamTunnelDash.cs`.

kfish610:
> My thought would be to allow a comma-separated list of classnames/entity IDs for mapmakers to input for DreamRefills... but that's not very usable for the global toggle.
Maybe if it's not modified by an entity then it uses the defaults you currently have set up?

This seems like the best option to me, though to be honest I'm not sure how that would work (not because it wouldn't, but because I'm a new modder, so I have no clue).

> Respects OnDashCollide and tunnels

I can see this being useful for buttons, but for Kevins it seems rather confusing, and I'm not sure what the effects would be on crumble blocks? There are also other things that could be done through IL, such as checking the DashCollision result and only tunnelling on certain types (though this would sadly require a special case for buttons), tho I'm not sure whether its worth doing too much IL given its ability to mess stuff up.

PR #94, opened 2022-01-19
An option for TimedTriggerSpikes to be direction independent when triggering

tkavner:


tkavner:
I forgot a description but it these changes let triggering Timed Trigger Spikes happen regardless of what direction the player is facing, allowing map creators to have players trigger them on jump throughs and the like.

These changes will not change the behavior of currently existing TimedTriggerSpikes (ie without this option enabled), nor will it change the outputs of any public methods for any preexisting TimedTriggerSpikes.

PR #99, opened 2022-03-27
Lönn Support

catapillie:
All the Ahorn plugins, written in Julia, were ported to Lua and made to function with Lönn. Open to feedback, as I'm sure this could be improved a lot!

There are also some addition and tweaks to tooltips, mostly filling up missing tooltips. Fixes some small bugs as well.


catapillie:
Post-draft changes:
* Gave dreamblock entities a custom rectangle border color for the `oneUse` option. That color is `Color.Firebrick` from XNA.
* Change dreambooster path sprites; node is shown at all times, with 25% opacity. The path is always shown, and made transparent when the `hidePath` option is on.
* Added missing indicator for floating dreamjellyfish.
* Made moveblockredirect `overrideColor` and `overrideUsedColor` color fields. They are hidden in all placements except in a new "reskinnable" one.
* Fixed melvin and chainedkevin entities generating newer different sprites often because of randomness. The randomness seed is set with `math.randomseed(x, y)` where `x` and `y` are the coordinates of the entity. It's good enough for entities that have an undetermined size, since the random numbers won't always fall on the same tiles.
* Fixed language file errors.

PR #107, opened 2022-07-16
Add DreamDashListener and DreamStrawberry

bunnerfly:
Someone in Celestecord posted a suggestion for a winged berry that is triggered on dream block enter. I thought this was a cool idea so I made it, also includes a DreamDashListener which was used for this which seems right up CommunalHelper's alley.

I don't have a general purpose public helper, so I thought this would be the best place for it!

Changes I made
- Added DreamDashListener
- Implemented DreamDashListener to DreamTunnelDash
- Added DreamStrawberry
- Added Ahorn Plugin for DreamStrawberry

Also, sorry about me being horrible at knowing when to make small commits. This whole thing just ended up as one commit lol


.gitignore
```diff
@@ -357,4 +357,5 @@ MigrationBackup/
 .ionide/
 
 # Fody - auto-generated XML schema
-FodyWeavers.xsd
\ No newline at end of file
+FodyWeavers.xsd
+/Maps/CommunalTestmap/test.bin
```
catapillie: Communal Helper doesn't have a test map, so I don't see why this is necessary.

src/Entities/Misc/DreamDashListener.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Celeste.Mod.CommunalHelper.Entities {
+
+    [Tracked(false)]
+    public class DreamDashListener : Component {
+
+        public Action<Vector2> OnDreamDash;
+
+        // This is actually unused in the base game, but exists in the normal DashListener, so it may as well be here.
+        public Action OnSet;
```
catapillie: I would rather remove it, since you have the possibility to do so, though this is debatable.

src/Entities/Misc/DreamStrawberry.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        /*
+         * Hey! snowii here
+         * 
+         * If anyone is reading this and wants to touch my unholy code AND has a sprite for this, please add it im begging
+         * 
+         * Hope someone finds this super niche excuse for an entity useful
+         */
+
+        private MethodInfo OriginalOnDash;
```
catapillie: This can be `private static readonly` instead of being an instance field, set for every Dream Berry entity.
In fact, you can directly initialize it as you declare it:
```cs
private static readonly MethodInfo m_Strawberry_OnDash = typeof(Strawberry).GetMethod("OnDash", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.InvokeMethod);
// Note: I renamed it, starting with an "m" for "method", followed by the type and then the name of the method. This isn't necessary, just a nitpick :)
```

src/Entities/Misc/DreamStrawberry.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        public static Color[] DreamTrailColors = new Color[] {
+            Calc.HexToColor("FFEF11"),
+            Calc.HexToColor("08A310"),
+            Calc.HexToColor("FF00D0"),
+            Calc.HexToColor("5FCDE4"),
+            Calc.HexToColor("E0564C")
+        };
+
+        public static DynamicData dreamStrawberrySpriteData;
```
catapillie: You probably don't want this to be `static`. Each time a Dream Stawberry is instantiated, you are setting this field to a value which takes the instance of that entity. So basically, everytime you access this field, it's the data of the latest instantiated Dream Strawberry, and not the one you actually want.
(Also the name is a bit misleading, I'd call it `strawberryData` instead)

src/Entities/Misc/DreamStrawberry.cs
```diff
@@ -0,0 +1,135 @@
...
+                    Audio.Play("event:/game/general/strawberry_pulse", Position);
+                    SceneAs<Level>().Displacement.AddBurst(Position, 0.6f, 4f, 28f, 0.1f);
+                } else {
+                    Audio.Play("event:/game/general/strawberry_pulse", Position);
+                    SceneAs<Level>().Displacement.AddBurst(Position, 0.6f, 4f, 28f, 0.2f);
+                }
+            }
+        }
+        */
+        #endregion
```
catapillie, line 94: I'm going to assume that this will be either used or removed, since you included a bunch of custom sprites in this PR- which by the way, thank you!

src/Entities/Misc/DreamStrawberry.cs
```diff
@@ -0,0 +1,135 @@
...
+            Vector2 scale = new Vector2(Math.Abs(berrySprite.Scale.X), berrySprite.Scale.Y);
+            TrailManager.Add(berry, scale, color);
+        }
+
+        // Actually creates the dream trail
+        public static void CreateDreamTrail(DreamStrawberry berry) {
+            CreateTrail(berry, DreamTrailColors[DreamTrailColorIndex]);
+            ++DreamTrailColorIndex;
+            DreamTrailColorIndex %= 5;
+        }
```
catapillie: These methods can probably be merged, and don't have to be marked `static`. Instead of passing the `DreamStrawberry` object as a parameter, don't make it static, and use directly reference the things you need from the instance (or with `this`), because that method now belongs to the `DreamStrawberry` instance.

catapillie (changes requested):
Ahorn plugin looks fine. It does have missing tooltips, which in this case aren't *super* important, but I would prefer if they were added. You can find the strawberry option tooltips [here](https://github.com/CelestialCartographers/Ahorn/blob/master/lang/en_gb.lang#L11).
This entity needs a Lönn plugin. If you're unfamiliar with those, I would recommend looking at [Lönn's strawberry plugin](https://github.com/CelestialCartographers/Loenn/blob/master/src/entities/strawberry.lua), copying it and modifying so that it only shows what you want it to show. Alternatively, I'll implement it myself.

.gitignore
bunnerfly: Not sure why it added my random test map in to the gitignore, next time ill probably just leave it in another map

src/Entities/Misc/DreamDashListener.cs
bunnerfly: I've left it in, but commented it out, I still think unused code is funny enough to not really mind

src/Entities/Misc/DreamStrawberry.cs
bunnerfly: I'm not sure why I didn't realize i could do that
also thank you for the mention on the naming, will do that for any other methods i have like this

src/Entities/Misc/DreamStrawberry.cs
bunnerfly: oh yeah that's a good point, will fix these too

src/Entities/Misc/DreamStrawberry.cs
bunnerfly: Merged the two methods and fixed them being static
Also swapped out the `DreamStrawberry` object to be referenced using `this`

catapillie (changes requested):
You've accidentally committed & pushed a zipped build of Communal Helper. Please remove it! Thanks for the other changes!

src/Entities/Misc/DreamStrawberry.cs
catapillie: Will resolve this for now, might work on the sprite on my own.

catapillie (commented):
Should be good, though @coloursofnoise may have a few things to say. Should be merged otherwise.

catapillie (approved):

Ahorn/entities/dreamDashBerry.jl
```diff
@@ -0,0 +1,25 @@
+module DreamDashListenerDreamDashBerry
+
+using ..Ahorn, Maple
+
+@mapdef Entity "CommunalHelper/DreamStrawberry" DreamDashBerry(x::Integer, y::Integer, order::Integer=-1, checkpointID::Integer=-1)
```
coloursofnoise: I don't know how well we've kept up with this, but it'd be good to run [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl) on this to have each property on a new line

Ahorn/entities/dreamDashBerry.jl
```diff
@@ -0,0 +1,25 @@
...
+function Ahorn.selection(entity::DreamDashBerry)
+	x, y = Ahorn.position(entity)
+
+	return Ahorn.getSpriteRectangle("collectables/strawberry/wings01", x, y)
+end
+
+function Ahorn.renderAbs(ctx::Ahorn.Cairo.CairoContext, entity::DreamDashBerry, room::Maple.Room)
+	x, y = Ahorn.position(entity)
+
+	Ahorn.drawSprite(ctx, "collectables/strawberry/wings01", x, y)
```
coloursofnoise: Despite it not being implemented in-game yet, should the plugins use the custom sprites provided? I assume we're going to want to add them to the in-game entity before releasing anyways

src/Entities/Misc/DreamDashListener.cs
coloursofnoise: I think it's best to remove this entirely, if it had some sort of meaning then I'd be fine with it but as it is it's just unnecessary and confusing

src/Entities/Misc/DreamStrawberry.cs
```diff
@@ -0,0 +1,130 @@
...
+            // I'd love to say I wrote this, but this is 100% stolen from SC2020 Source
+            // Whoever wrote this piece of code so I do not have to, you're my bestie
+
+            //OriginalOnDash = typeof(Strawberry).GetMethod("OnDash", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.InvokeMethod);
+
+            // Removes any default DashListeners from the strawberry as we do not want to use those
+            foreach (Component comp in Components.ToArray()) {
+                if (comp is DashListener)
+                    Components.Remove(comp);
+            }
```
coloursofnoise, line 44:
```suggestion
            Components.RemoveAll<DashListener>();
```

src/Entities/Misc/DreamStrawberry.cs
```diff
@@ -0,0 +1,130 @@
...
+        public DynamicData dreamStrawberryData;
+
+        public static int DreamTrailColorIndex = 0;
+
+        public DreamStrawberry(EntityData data, Vector2 offset, EntityID id) : base(FixData(data), offset, id) {
+
+            // I'd love to say I wrote this, but this is 100% stolen from SC2020 Source
+            // Whoever wrote this piece of code so I do not have to, you're my bestie
+
+            //OriginalOnDash = typeof(Strawberry).GetMethod("OnDash", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.InvokeMethod);
```
coloursofnoise: What code is this comment referring to? the `OriginalOnDash` assignment should be removed as it's unused

src/Entities/Misc/DreamStrawberry.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            // Creates and updates the dream trail
+            if (Scene.OnInterval(0.1f))
+                CreateDreamTrail();
+        }
+
+        // Calls the original OnDash from our DreamDashListener
+        public void OnDreamBerryDash(Vector2 dir) {
+            m_Strawberry_OnDash.Invoke(this, new object[] { dir });
+        }
```
coloursofnoise, line 108: I wonder if this could be handled with a `MonoModLinkTo`? something like 
```suggestion
        // Calls the original OnDash from our DreamDashListener
        [MonoModLinkTo("Celeste.Strawberry", "OnDash(Microsoft.Xna.Framework.Vector2)")]
        public extern void OnDreamBerryDash(Vector2 dir);
```
no clue if that actually works but it'd be a really clean way to do this if it does

coloursofnoise (changes requested):
Looks good overall, just some extra cleanup and questions.

Reminder to whoever merges this to do a Squash Merge to make sure the `CommunalHelperfuny.zip` file doesn't get included in the repo.

src/Entities/Misc/DreamStrawberry.cs
bunnerfly: well yeah in theory this would be a clean solution, but this implementation causes the berry to refuse to load, so i think the current implementation is good enough for now.

i dont really see a need to replace it either way

src/Entities/Misc/DreamStrawberry.cs
bunnerfly: it was referring to the commented out code, which is now removed alongside the comment

src/Entities/Misc/DreamStrawberry.cs
bunnerfly: this too would be a lot better of a method to do this, but for whatever reason this likes the throw a TargetInvocationException

Ahorn/entities/dreamDashBerry.jl
bunnerfly: Good idea, not sure why I didn't think of this yet
Implemented in both Ahorn and Loenn

Ahorn/entities/dreamDashBerry.jl
bunnerfly: Fixed up the properties being a mess

src/Entities/Misc/DreamStrawberry.cs
coloursofnoise: As discussed on discord, the `RemoveAll` implementation will need to be fixed in Everest before this is an option

coloursofnoise (approved):

PR #112, opened 2022-07-26
Implement Curved Boosters

catapillie:
image
These were implemented for quite a bit. The non-dream variant won't allow you to go through walls. This branch also does some refactoring with the `CustomBooster` & `DreamBooster` classes.


catapillie:
Need to add GravityHelper support to curved boosters. They don't work with inverted gravity.

PR #113, opened 2022-07-26
Add DreamStrawberry custom sprite & seeds

catapillie:
The sprites included for the Dream Strawberry were edited and used with an IL hook. Dream Strawberries can also have seeds, which themselves behave like vanilla seeds.
dreamberry idle
dreamberry idle


catapillie:
Add custom Dream Strawberry Seed behavior: you lose them when you grab or stand on a non-dreamblock solid.

PR #117, opened 2022-08-25
Removed some sources of log spam

bigkahuna443:
1. Unnecessary animations with missing images for Redless Berry
2. A texture access with fallback vs. just testing for key existence in ConnectedMoveBlock (fallback usage could be further cleaned up, but this removes the log spam)
3. Hook logging which should probably be lower priority in AbstractPanel

### DIFF
--- a/src/Entities/Panels/AbstractPanel.cs
+++ b/src/Entities/Panels/AbstractPanel.cs
@@ -158,7 +158,7 @@ namespace Celeste.Mod.CommunalHelper.Entities {
 
         private static void LogHooked(string fullname) {
             if (!hooked.Contains(fullname)) {
-                Logger.Log(LogLevel.Verbose, "CommunalHelper", $"Hooking {fullname} to override when AbstractPanel present.");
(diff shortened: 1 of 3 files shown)


catapillie (approved):
Thanks a lot, looks alright to me.

src/Entities/Panels/AbstractPanel.cs
```diff
@@ -158,7 +158,7 @@ public AbstractPanel(Vector2 position, float size, Directions orientation, bool
 
         private static void LogHooked(string fullname) {
             if (!hooked.Contains(fullname)) {
-                Logger.Log(LogLevel.Info, "CommunalHelper", $"Hooking {fullname} to override when AbstractPanel present.");
+                Logger.Log(LogLevel.Verbose, "CommunalHelper", $"Hooking {fullname} to override when AbstractPanel present.");
```
coloursofnoise: I think I'd prefer for this to be at least `Debug`, since it's something that I worry about causing weird conflicts for other mods

bigkahuna443: Yeah that makes sense, that's what we changed detour logs to anyway (although you could still use the command to print them out regardless). Updated.

PR #118, opened 2022-09-04
Add ModInterop exports

coloursofnoise:
Makes it easier for other mods to integrate with communalhelper stuff without having to explicitly reference CommunalHelper

### DIFF
--- a/src/ModInterop.cs
+++ b/src/ModInterop.cs
@@ -16,7 +16,7 @@ namespace Celeste.Mod.CommunalHelper {
 
             #region DreamTunnel
 
-            public static int StDreamTunnelDash() => DreamTunnelDash.StDreamTunnelDash;
+            public static int GetDreamTunnelDashState() => DreamTunnelDash.StDreamTunnelDash;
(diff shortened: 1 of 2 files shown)


src/ModInterop.cs
```diff
@@ -0,0 +1,37 @@
...
+        internal static void Initialize() {
+            typeof(DashStates).ModInterop();
+        }
+
+        [ModExportName("CommunalHelper.DashStates")]
+        public static class DashStates {
+
+            #region DreamTunnel
+
+            public static int StDreamTunnelDash() => DreamTunnelDash.StDreamTunnelDash;
```
catapillie: As mentioned on Discord, a name like `GetDreamTunnelDashState` or `GetStDreamTunnelDash` would make a lot more sense for this. I personally prefer `GetDreamTunnelDashStateID`, I think it's as clear as you can get.

catapillie (commented):
Apart from an export's name that could be improved, pretty much good otherwise.

bigkahuna443 (approved):

PR #119, opened 2022-09-08
Fix Scrape Particle Interval

bigkahuna443:
There were a couple Connected Solid types that didn't have the scrape particle interval set so they would spam particles.

Zip mover particle quantity is definitely higher than it should be + the direction isn't correct with the default, that can be looked at later though :) probably shouldn't share the same logic


catapillie (approved):
thanks :)

PR #122, opened 2022-09-11
implement connected temple cracked blocks

corkr900:
Connected version of temple cracked blocks. They are broken by seeker charge, seeker regen, puffer explosion, and player seeker charge. Persistent option. Not currently reskinnable but coded in a way that would make that pretty easy to implement in case anyone ever wants it in the future. Entities from this or any other mod that should break them (e.g. cavern crystal bombs) will have to implement support themselves as there is no general way to do it from the block's side.

Loenn support not included at the time of creating the PR because i have no clue how to do that.


Ahorn/entities/connectedBlocks/ConnectedTempleCrackedBlock.jl
```diff
@@ -0,0 +1,70 @@
...
+end
+
+function Ahorn.render(ctx::Ahorn.Cairo.CairoContext, entity::ConnectedTempleCrackedBlock, room::Maple.Room)
+    x = Int(get(entity.data, "x", 0))
+    y = Int(get(entity.data, "y", 0))
+
+    width = Int(get(entity.data, "width", 32))
+    height = Int(get(entity.data, "height", 32))
+
+    rendertempleCrackedBlock(ctx, x, y, width, height)
```
catapillie, line 67: You did take care of the Lönn plugin, which is awesome, but you forgot to update the rendering logic for the Ahorn plugin. It would be awesome if the two plugins could display similarly in both Ahorn and Lönn.
You can take a look at the [ConnectedZipMover.jl](https://github.com/CommunalHelper/CommunalHelper/blob/dev/Ahorn/entities/connectedBlocks/ConnectedZipMover.jl) plugin to see how to do the same, but in Julia for Ahorn.

_(In reality though, this isn't that much of a issue, since we're not really supporting Ahorn anymore)._

src/Entities/ConnectedStuff/ConnectedTempleCrackedBlock.cs
```diff
@@ -0,0 +1,280 @@
...
+using MonoMod.Utils;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Celeste.Mod.CommunalHelper.Entities.ConnectedStuff {
```
catapillie: All Communal Helper entities reside in the `Celeste.Mod.CommunalHelper.Entities` namespace.
```suggestion
namespace Celeste.Mod.CommunalHelper.Entities {
```

src/CommunalHelperModule.cs
```diff
@@ -52,6 +52,8 @@ public class CommunalHelperModule : EverestModule {
             DreamBlockDummy.Load();
 
             CustomDreamBlock.Load();
+            Entities.ConnectedStuff.ConnectedTempleCrackedBlock.Load();
```
catapillie: Once you change the namespace for the `ConnectedTempleCrackedBlock` class, you can apply this change:
```suggestion
            ConnectedTempleCrackedBlock.Load();
```
(Same for the `ConnectedTempleCrackedBlock.Unload` call).
You can also move it up with the other connected-type entities if you'd like (looks a bit cleaner).
Yeah, this is just a small nitpick :)

Loenn/lang/en_gb.lang
```diff
@@ -332,6 +332,11 @@ entities.CommunalHelper/ConnectedMoveBlock.attributes.description.pressedColor=T
 entities.CommunalHelper/ConnectedMoveBlock.attributes.description.breakColor=The color of the arrow when breaking.
 entities.CommunalHelper/ConnectedMoveBlock.attributes.description.outline=Gives this Move Block an outline.
 
+# Connected Temple Cracked Block
+entities.CommunalHelper/ConnectedTempleCrackedBlock.placements.name.normal=Connected Temple Cracked Block
+entities.CommunalHelper/ConnectedTempleCrackedBlock.placements.name.persistent=Connected Temple Cracked Block (Persistent)
+entities.CommunalHelper/ConnectedTempleCrackedBlock.attributes.description.persistent=Persistent blocks will stay broken after dying or leaving the room.
```
catapillie, line 338: You forgot to add this tooltip to the Ahorn `en_gb.lang` file.

catapillie (changes requested):
Overall pretty good stuff, some small things need to be changed, and the Ahorn plugin could be updated.

Ahorn/entities/connectedBlocks/ConnectedTempleCrackedBlock.jl
corkr900: I'm not sure what you mean here. They appear almost identically in loenn + ahorn for me, with only a minor tiling difference where it meets extensions that doesnt seem important to me since we're phasing out ahorn anyway.
Is there another problem that I'm overlooking?

Ahorn:
screenshot-1 png (139)

Loenn:
screenshot-1 png (140)

Loenn/lang/en_gb.lang
corkr900: What tooltip? The only attribute besides position/size is the persistent checkbox and that's already in there

src/Entities/ConnectedStuff/ConnectedTempleCrackedBlock.cs
corkr900: Updated.

src/CommunalHelperModule.cs
corkr900: Updated. Did not move since there did not seem to be a clear grouping for it to go in (most of the other connected entities do not have a function called during Load() )

Ahorn/entities/connectedBlocks/ConnectedTempleCrackedBlock.jl
catapillie: Yeah, that's fine. It's not as bad as I thought it would be.

src/CommunalHelperModule.cs
catapillie: No worries about it then. The `Load` & `Unload` calls can be moved afterwards anyway, during manual cleanup for example.

Loenn/lang/en_gb.lang
corkr900: OH WAIT you said Ahorn not Loenn. I'll go do that...

catapillie: I can't find it. The PR doesn't list any change in `Ahorn/lang/en_gb.lang` either.

corkr900: ok done

catapillie (approved):

PR #124, opened 2022-09-30
fix rubble spawning in connected temple cracked blocks

corkr900:
Previously, rubble from breaking a connected temple cracked block resulted in a massive rectangle of rubble that didn't even necessarily align with the connected block's extensions.

With this fix, rubble spawns at each tile that was occupied by the source block or a collidable extension.

Video credit to bigkahuna:

https://user-images.githubusercontent.com/109830893/193165101-53cdd2e6-a257-45f0-a86c-0e18a32e1cfe.mp4


bigkahuna443 (approved):
Tested, looks good to me 👍

PR #127, opened 2022-11-08
add shielded refill lönn plugin

microlith57:
this is the only remaining ahorn plugin without a corresponding lönn plugin


catapillie (approved):
looks good, thanks!

PR #128, opened 2022-11-28
Improve Attached Wall Booster core mode handling

Kalobi:
The `notCoreMode` option inherited from vanilla has very idiosyncratic behavior (it forces the wall booster to be in ice mode when entering the room, but doesn't prevent it from toggling with the core mode). This PR retires that option (retaining backwards compatibility) and replaces it with a dropdown that offers the options of always using the hot/cold variant or toggling based on core mode, and also allows specifying in that case which mode to use if the core mode is None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -358,3 +358,6 @@ MigrationBackup/
 
 # Fody - auto-generated XML schema
 FodyWeavers.xsd
+
+# JetBrains Rider
+.idea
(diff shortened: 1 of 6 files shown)


catapillie (approved):
Looking all good, thanks a lot. There's a funny `switch` expression that looks weird, but that's fine. Also thanks for cleaning up lang files and improving the code overall :)

Issue #121, opened 2022-09-10
Connected Move Blocks Custom Audio Parameter


vivianlonging:
**Describe your request:**
Move SFX Play overridden with stored string event value, pretty simple.
Not urgent
Request by TyKim

PR #140, opened 2023-05-27
Added support to overide SFX in Connected Move Blocks

mayromr:
Added support to overide SFX in Connected Move Blocks as requested by #121.
I was not sure if I should add a Lonn field for each sound effect or just a prefix, so I went with the prefix.

And a general question, do you have to compile this mode against the `opengl` version of Celeste? I faileds to compile it aganisnt the noraml game version.


catapillie (changes requested):
Hey, thanks for looking into this. While I appreciate that you implemented the custom sound option, I'm unsure whether it was really necessary to define a new class to handle custom paths. In my opinion, storing four `protected string` fields (one for each sound effect) works just as well.

In the constructor, all you need to do is get the custom path from entity data, trim unwanted characters, also insert `event:/` if it's not already there (like you already do, sure), and then manually append `_activate`, `_break`, `_reform_begin`, `_reappear` the the custom path to get the sfx event names. You still need to check whether they are valid paths, so that you can fall back the to default vanilla ones in case they aren't valid, but you've already got that covered in `DynamicSoundEffect.TryLoadNewSoundEffect` so that's good!

So, to sum up, I'd much prefer if you moved everything to the constructor, or to a method (maybe `LoadCustomSounds` which you call from the constructor:
```cs
// @@ ConnectedMoveBlock
protected string ActivateSoundEffect = SFX.game_04_arrowblock_activate;
protected string BreakSoundEffect = SFX.game_04_arrowblock_break;
protected string ReformBeginSoundEffect = SFX.game_04_arrowblock_reform_begin;
protected string ReappearSoundEffect = SFX.game_04_arrowblock_reappear;
```
and then
```cs
// @@ public ConnectedMoveBlock(EntityData data, Vector2 offset)
string customSoundEffectPath = data.Attr("customSoundEffect").Trim().TrimEnd('/').TrimEnd('_');
if (!string.IsNullOrWhiteSpace(customSoundEffectPath))
{
    if (!customSoundEffectPath.StartsWith("event:/"))
    {
        customSoundEffectPath = customSoundEffectPath.TrimStart('/');
        customSoundEffectPath = $"event:/{customSoundEffectPath}";
    }

    // create _activate sound path
    // make sure it exists, and it if does, store in in `this.ActivateSoundEffect`
    
    // do the same for all the three other sound effects (_break, _reform_begin, _reappear).
}
```

(Note that, with your current implementation, the `_reform_begin` is actually just referred to as `_begin`, since you use the last underscore's index).

For building the mod, I _think_ you need to have the mod's repository cloned in your Celeste's `Mods` folder (that's how the .csproj file is setup AFAIK). Make sure you're running the FNA version of the game, not XNA (you may already have it installed, depending on where you got the game from. On Epic Games Store, it's FNA. on Steam, you need to select it, etc...)

Feel free to ask any questions, hope to hear back.

mayromr:
One quetion about the logs, I removed them because I saw duplicates logs for each missing SFX, it seems that `Audio.GetEventDescription` already logs about a missing SFX. Is that okay or should I readd them?

catapillie:
> One quetion about the logs, I removed them because I saw duplicates logs for each missing SFX, it seems that `Audio.GetEventDescription` already logs about a missing SFX. Is that okay or should I readd them?

You don't need to add them back, it's fine as is.

Will check out the new changes when I get back at home!

catapillie (approved):
Looks good, thanks a lot. Will very likely merge later today

PR #141, opened 2023-07-01
Use DynamicData instead of DynData

catapillie:
Replaces every usage of `DynData<T>` with `DynamicData`, and replaces `DynData.this[string]` indexing with `DynamicData.Get` & `DynamicData.Set` calls.


catapillie (commented):
Because `DynData<T>` is a generic type, I've checked that every `Dynamic.For` call gets passed an an object of the corresponding type in the previous implementation. Otherwise, I use `new DynamicData(typeof(...), obj)`

PR #146, opened 2023-07-13
Add everything that came from 3D objects

catapillie:
3D objects, in MY Celeste?

This PR contains all the stuff that was created with 3D objects. 3D objects are stored with `Mesh<VertexPCTN>`, and drawn by a custom renderer entity with performs deferred rendering. PCTN stands for "Position Color Texture Normal", and I've just named everything after it. The rendering has three phases (two shaders + extra draw call):
1. Deferred rendering of the geometry onto the albedo (color), depth and normal buffers.
2. Composition of the final image.
3. The final image is drawn onto the Gameplay buffer.

3D meshes are wrapped into a component which can be added to entities, `Shape3D`. It contains various fields to tweak the appearance of the model, like outlines, highlights, rainbow, tint, etc... I won't be going into what they each do.

New entities:
* Flying Aero Blocks - hovering blocks that follow a sequence of nodes, not useful for other than decoration/scenery (?)
* Charged Aero Block - the same as the above, but carries buttons on its sides which give boosts to the player when jumping/wall-jumping/climb-jumping/wallbouncing/hyperdashing/superdashing off of them.
* Shapeshifter - looks like a solid 2D tilemap until it rotates in 3d and re-aligns with the grid to change shape and become another 2D tilemap. Probably the most high effort yet most useless thing I've ever coded in here.
* Elytra Rings - 3D rings that the player can traverse to do various stuff: speed boost, refill, stop, or play music (funny).

Adds a tool, the Voxel Editor, to make the 3D voxel shapes (for shapeshifter).


catapillie:
finally done

PR #147, opened 2023-07-19
Add new Falling and Pushable components

swoolcock:
Should probably include a sample entity too.

### DIFF
--- a/src/Components/Falling.cs
+++ b/src/Components/Falling.cs
@@ -85,13 +85,13 @@ public class Falling : Component
     private void OnShakeSfx()
     {
         if (Entity is not null)
-            Audio.Play("event:/game/general/fallblock_shake", Entity.Center);
+            Audio.Play(SFX.game_gen_fallblock_shake, Entity.Center);
     }
 
     private void OnImpactSfx()
     {
         if (Entity is not null)
-            Audio.Play("event:/game/general/fallblock_impact", Entity.BottomCenter);
+            Audio.Play(SFX.game_gen_fallblock_impact, Entity.BottomCenter);
     }
 
     private void OnFallParticles()
--- a/src/Components/Falling.cs
+++ b/src/Components/Falling.cs
@@ -1,0 +1,258 @@
+using System.Collections;
+
+namespace Celeste.Mod.CommunalHelper.Components;
+
+public class Falling : Component
+{
+    // actions
+    public Func<bool> PlayerFallCheck;
+    public Func<bool> PlayerWaitCheck;
+    public Action ShakeSfx;
+    public Action ImpactSfx;
+    public Action LandParticles;
+    public Action FallParticles;
+
+    // config
+    public float FallDelay;
+    public bool ShouldRumble = true;
+    public bool ClimbFall = true;
+    public float FallSpeed = 160f;
+    public bool ShouldManageSafe = true;
+    public bool FallUp = false;
+    public float ShakeTime = 0.4f;
+    public bool EndOnSolidTiles = true;
+    public bool FallImmediately = false;
+
+    // coroutine properties
+    public bool Triggered;
+    public bool HasStartedFalling { get; private set; }
+
+    private Coroutine coroutine;
+
+    public new Solid Entity => EntityAs<Solid>();
+
+    public Falling() : base(false, false)
+    {
+        PlayerFallCheck = OnPlayerFallCheck;
+        PlayerWaitCheck = OnPlayerWaitCheck;
+        ShakeSfx = OnShakeSfx;
+        ImpactSfx = OnImpactSfx;
+        LandParticles = OnLandParticles;
+        FallParticles = OnFallParticles;
+    }
+
+    public override void Added(Entity entity)
+    {
+        // need to call base first to ensure we balance Add/Remove
+        base.Added(entity);
+
+        if (entity is not Solid)
+        {
+            Logger.Log(LogLevel.Warn, "CommunalHelper", $"Attempted to add {nameof(Falling)} to a non-Solid ({entity.GetType().Name})");
+            RemoveSelf();
+            return;
+        }
+
+        Reset();
+    }
+
+    public void Reset()
+    {
+        coroutine?.RemoveSelf();
+        Entity?.Add(coroutine = new Coroutine(FallingSequence()));
+        if (ShouldManageSafe && Entity is not null)
+            Entity.Safe = false;
+    }
+
+    private bool OnPlayerFallCheck()
+    {
+        if (Entity is null)
+            return false;
+        return ClimbFall ? Entity.HasPlayerRider() : Entity.HasPlayerOnTop();
+    }
+
+    private bool OnPlayerWaitCheck()
+    {
+        if (Entity is null)
+            return false;
+        if (Triggered || PlayerFallCheck?.Invoke() == true)
+            return true;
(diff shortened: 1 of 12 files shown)


src/Components/Falling.cs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    public override void Added(Entity entity)
+    {
+        // need to call base first to ensure we balance Add/Remove
+        base.Added(entity);
+
+        if (entity is not Solid)
+        {
+            Logger.Log(LogLevel.Warn, "CommunalHelper", $"Attempted to add {nameof(Falling)} to a non-Solid ({entity.GetType().Name})");
```
catapillie: can use `Util.Log` instead so you don't have to specify `"CommunalHelper"`

src/Components/Falling.cs
```diff
@@ -0,0 +1,258 @@
...
+    {
+        if (Entity is not null)
+            Audio.Play("event:/game/general/fallblock_shake", Entity.Center);
+    }
+
+    private void OnImpactSfx()
+    {
+        if (Entity is not null)
+            Audio.Play("event:/game/general/fallblock_impact", Entity.BottomCenter);
+    }
```
catapillie: Can use the constants strings in `Celeste.SFX`

src/Components/Falling.cs
```diff
@@ -0,0 +1,258 @@
...
+            // if it's hit the fg tiles then make it safe and end
+            if (EndOnSolidTiles && entity.CollideCheck<SolidTiles>(entity.Position + (FallUp ? -Vector2.UnitY : Vector2.UnitY)))
+            {
+                entity.Safe |= self.ShouldManageSafe;
+                yield break;
+            }
+
+            // wait until we can fall again
+            while (entity.CollideCheck<Platform>(entity.Position + (FallUp ? -Vector2.UnitY : Vector2.UnitY)))
+                yield return 0.1f;
```
catapillie, line 255: Is there a reason for this specific wait time? Would `yield return null` work? Or maybe that's how vanilla works.

src/Components/Pushable.cs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    public override void Added(Entity entity)
+    {
+        // need to call base first to ensure we balance Add/Remove
+        base.Added(entity);
+
+        if (entity is not Solid)
+        {
+            Logger.Log(LogLevel.Warn, "CommunalHelper", $"Attempted to add {nameof(Pushable)} to a non-Solid ({entity.GetType().Name})");
```
catapillie: `Util.Log`

catapillie (commented):
Looks great so far, except some nitpicks :)

src/Components/Falling.cs
swoolcock: Didn't see that, I just did a search for `Logger.Log` and and found some instances of it, so I used that.

src/Components/Falling.cs
swoolcock: Yeah this is how vanilla does it in `FallingBlock`:
```cs
while (fallingBlock.CollideCheck<Platform>(fallingBlock.Position + new Vector2(0.0f, 1f)))
    yield return (object) 0.1f;
```

catapillie: okay, nevermind then!

catapillie (commented):
i'll approve the pr once we have a working entity

swoolcock:
Yeah I've started with the slingshot block, will get back to it tomorrow.

src/Entities/Aero/AeroBlockSlingshot.cs
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    private int IdleUpdate()
+    {
+        return (int) SlingshotStates.Idle;
+    }
+
+    private void IdleEnd()
+    {
+    }
```
catapillie, line 160: AFAIK It is possible to remove those and pass `null` to `StateMachine.SetCallbacks` where they are passed. But honestly I don't mind if they're left here for clarity

src/Entities/Aero/AeroBlockSlingshot.cs
```diff
@@ -0,0 +1,417 @@
...
+        {
+            Position = new Vector2(width, height) / 2.0f,
+        });
+
+        stateMachine = new StateMachine();
+        stateMachine.SetCallbacks((int) SlingshotStates.Idle, IdleUpdate, begin: IdleBegin, end: IdleEnd);
+        stateMachine.SetCallbacks((int) SlingshotStates.Windup, WindupUpdate, begin: WindupBegin, end: WindupEnd);
+        stateMachine.SetCallbacks((int) SlingshotStates.Locked, LockedUpdate, begin: LockedBegin, end: LockedEnd);
+        stateMachine.SetCallbacks((int) SlingshotStates.Launch, LaunchUpdate, begin: LaunchBegin, end: LaunchEnd);
+        stateMachine.SetCallbacks((int) SlingshotStates.Cooldown, CooldownUpdate, begin: CooldownBegin, end: CooldownEnd);
```
catapillie, line 129: `StateMachine` has a `IEnumerator` callback for each state, which can be used as a regular coroutine. For states that cannot be interrupted, like Launch and Cooldown, we could use the coroutine callbacks and `yield return Util.Interpolate` like we used to, instead of relying on a timer field

catapillie (commented):
Looks really good, though I wonder if we can use coroutines in the state machine in some places instead of an update function which relies on a timer.
I'll mark the PR as ready, and will approve the changes when we decide on if there is anything we want to change.

src/Entities/Aero/AeroBlockSlingshot.cs
```diff
@@ -347,6 +348,9 @@ private void OnPush(int moveX, Pushable.MoveActionType moveAction)
     {
         stateMachine.State = (int) SlingshotStates.Windup;
         updateTimer = SetTime;
+
+        var speed = (pushable.MaxPushSpeed < 0 ? 70f : Math.Min(Player.MaxRun, pushable.MaxPushSpeed)) * Input.MoveX.Value;
```
swoolcock: This should use `moveX` rather than `Input.MoveX.Value`, otherwise it's restricted to actual player input.  In the future we may want to other entities to interact with `Pushable`s.

`OnPush` is already called using `Input.MoveX.Value`, but not relying on this outside the component gives us greater flexibility in the future.

src/Entities/Aero/AeroBlockSlingshot.cs
swoolcock: Yeah I put all the stubs in initially in case I needed them, but I forgot to go through and remove them afterward.

src/Entities/Aero/AeroBlockSlingshot.cs
swoolcock: I looked at using a per-state coroutine but I just wanted it to be done.  Maybe a future task.

src/Entities/Aero/AeroBlockSlingshot.cs
catapillie: gotcha. just pushed the change. should've been a little more careful instead of pasting the line there

catapillie:
alright. merging now. thanks a lot for the components, as well for this entity. i'm actually super glad with how that turned out!!! thanks again

PR #149, opened 2023-09-02
Add Badeline Boost (Keep Holdables)

SnipUndercover:
I saw somebody ask for a Badeline Boost which didn't force you to drop a holdable and decided to implement it myself.

Tested it and it seems to work as intended.

### DIFF
--- a/src/Entities/BadelineBoosts/BadelineBoostKeepHoldables.cs
+++ b/src/Entities/BadelineBoosts/BadelineBoostKeepHoldables.cs
@@ -1,0 +1,50 @@
+﻿using Mono.Cecil.Cil;
+using MonoMod.Cil;
+using MonoMod.RuntimeDetour;
+using MonoMod.Utils;
+using System.Reflection;
+
+namespace Celeste.Mod.CommunalHelper.Entities.BadelineBoosters;
+
+[CustomEntity("CommunalHelper/BadelineBoostKeepHoldables")]
+public class BadelineBoostKeepHoldables : BadelineBoost
+{
+    private static readonly MethodInfo m_BoostRoutine =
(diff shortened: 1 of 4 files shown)


src/Entities/BadelineBoosts/BadelineBoostKeepHoldables.cs
```diff
@@ -0,0 +1,50 @@
+﻿using Mono.Cecil.Cil;
+using MonoMod.Cil;
+using MonoMod.RuntimeDetour;
+using MonoMod.Utils;
+using System.Reflection;
+
+namespace Celeste.Mod.CommunalHelper.Entities.BadelineBoosters;
```
catapillie: I'm not sure whether a new namespace + sub-directories (both in C# project and Lönn plugins) were necessary. Also, we usually limit namespaces to `Celeste.Mod.CommunalHelper.Entities`. Yes there are some entities for which we forgot to do this, and we can't change it back because it may break maps

catapillie (approved):
Looking good otherwise, thanks. I don't really mind the new folder all that much.

src/Entities/BadelineBoosts/BadelineBoostKeepHoldables.cs
SnipUndercover: The extra `BadelineBoosters` namespace is an oversight on my part, oops. I can remove that once I have access to VS again.

catapillie: NP; that'd be great, thanks!

catapillie (approved):
Thanks. Will merge at a later time, to make sure this is OK

PR #151, opened 2023-09-05
Fix Dash State Trigger

Kalobi:
Dash State Trigger wasn't loaded because the CustomEntity attribute only looks for public constructors.


Kalobi:
Pulling is a skill issue.

PR #152, opened 2023-09-05
Clean up DynamicData uses

Kalobi:
Make sure all DynamicData uses have the correct type (prevents MonoMod Crime warnings on core).
Remove unused DynamicData objects (DynamicData creation is expensive).


catapillie:
Thanks a lot!

PR #153, opened 2023-09-10
connected dreamblock + red booster fix

vivianlonging:
we are doing this again because it's not working rn. This could have originally been a modconflict scenario.

### DIFF
--- a/src/Entities/DreamBlocks/ConnectedDreamBlock.cs
+++ b/src/Entities/DreamBlocks/ConnectedDreamBlock.cs
@@ -109,7 +109,6 @@ public class ConnectedDreamBlock : CustomDreamBlock
     public bool HasGroup { get; private set; }
 
     public bool MasterOfGroup { get; private set; }
-
     public Dictionary<Platform, Vector2> Moves;
     public List<ConnectedDreamBlock> Group;
     public List<JumpThru> JumpThrus;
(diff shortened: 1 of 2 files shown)


catapillie:
Thanks, but how does this work? Looks like you just removed a DashCollision callback

vivianlonging:
The reason that this works is that, quoting you:
> re: that's not true. for a dream block to be dreamdashable, OnDashCollide must either be null or return DashCollisionResults.NormalOverride
This _does not_ work when the player is in a CustomDashState or in Red Boost state, therefore I'm setting it to be null.

If it makes more sense, I can implement a system where Update sets OnDashCollide to the previous behavior a la
```cs
if(PlayerHasDreamDash) OnDashCollide = null;
else OnDashCollide = (p,v) => return DashCollisionResults.NormalCollision;
```
(except I would personally suggest swapping the PlayerHasDreamDash function entirely except idk what's going to happen to the custom dream blocks)

catapillie:
Oh yeah, I remember. But now, because `OnDash` and `OnDashJumpThru` are `protected virtual`, isn't that going to cause issues for child classes?

vivianlonging:
Sorry, correction:
changing OnDash/OnDashJumpThru to that delegate and then setting it to OnDash
if it seems like the only thing we can do to 100% ensure OnDashCollide enables dreamdashes is to set it to null, (the bug this fixes being vertical dashcollide breaks for nondashstate), then there's not a whole lot we can do about that with

PR #160, opened 2023-10-05
Initial port of HintController from SJ

swoolcock:
Mostly the same as the one from Strawberry Jam, difference being that hint dialog is now selected with a counter instead of a flag, allowing for more than two hints.  It can also optionally cycle through these automatically.


swoolcock:
Should be good to go.  It doesn't touch any other entities etc. so if there's an issue with it we can fix it as required.

Issue #162, opened 2023-10-09
Add alpha to Cloudscapes


grog398:
**Describe your request:**

Add a way to change the transparency of the clouds/background for the Cloudscapes styleground, instead of just having an on/off checkbox for the background.

- [x] Is your request a change to an existing feature?
  - Cloudscapes
- [ ] Is this something that will be used for an upcoming collab/contest?

**Additional details:**
None

PR #165, opened 2023-10-17
Add alpha to Cloudscapes

SunsetQuasar:
`colors` and `bgColor` now have RGBA format support, which allows for separate cloud ring and background alpha control;
New `alpha` param controls the alpha of the Cloudscape's `VirtualRenderTarget`;
Implemented for Issue #162

### DIFF
--- a/Loenn/effects/cloudscape.lua
+++ b/Loenn/effects/cloudscape.lua
@@ -48,13 +48,17 @@ cloudscape.fieldInformation = {
     rotationExponent = {
         minimumValue = 0.0,
         maximumValue = 2.0
+    },
+    alpha = {
+        minimumValue = 0.0,
+        maximumValue = 1.0
     }
 }
 
 cloudscape.defaultData = {
     seed = "",
-    colors = "6d8ada,aea0c1,d9cbbc",
-    bgColor = "4f9af7",
+    colors = "6d8adaff,aea0c1ff,d9cbbcff",
+    bgColor = "4f9af7ff",
     innerRadius = 40.0,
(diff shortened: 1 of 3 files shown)


SunsetQuasar:
oops uhh lemme check if the code actually works fine!

SunsetQuasar:
nvm

src/Utils/Util.cs
```diff
@@ -57,6 +57,36 @@ public static Color TryParseColor(string str, float alpha = 1f)
         return CopyColor(Calc.HexToColor(str.Trim('#')), alpha);
     }
 
+    public static Color HexToColorWithAlphaNonPremultiplied(string hex)
```
microlith57, line 60: is it necessary to use nonpremultiplied colours? using premultiplied ones would be more consistent with everest decals, and i don't think having loads of different colour formats is a good idea

Loenn/effects/cloudscape.lua
```diff
@@ -48,13 +48,17 @@ cloudscape.fieldInformation = {
...
+        maximumValue = 1.0
     }
 }
 
 cloudscape.defaultData = {
     seed = "",
-    colors = "6d8ada,aea0c1,d9cbbc",
-    bgColor = "4f9af7",
+    colors = "6d8adaff,aea0c1ff,d9cbbcff",
+    bgColor = "4f9af7ff",
```
microlith57, line 61: ~~you've changed the defaults here but not the fallbacks in `Cloudscape.cs`; ideally those should match~~

PR #166, opened 2023-10-18
Fix crush block redirectable

mayromr:
This PR  contains 3 bug fixes:
1. a`MoveBlockRedirctable` constructor was using a null value
2. `MoveSwapBlock` was activating while broken from a `MoveBlockRedirect`
3. `connectedMoveBlock` not breaking on `MoveBlockRedirect`

It seems that `MoveSwapBlock` is still buggy with `MoveBlockRedirect`. If `MoveBlockRedirect` is set to slow redirect and the `MoveBlockRedirect` direction is perpendicular to the `MoveSwapBlock` direction, the game could crash.


vivianlonging:
Going to merge this since it's a solid fix for the time being and I'll handle the MoveSwapBlock in the next update because people are waiting on the update

PR #171, opened 2023-11-16
Made Elytra's animation more stable, and reskinable for skin

AAA1459:
Elytra's code seems badly with KyaruMod v1.3 or EeveeSkin v2.0 or my SMH Plus (just maybe), and cannot reskin for skin.
I made this to solve them.


src/States/Elytra.cs
```diff
@@ -436,12 +436,6 @@ private static PlayerDeadBody Mod_Player_Die(On.Celeste.Player.orig_Die orig, Pl
     private static void Mod_PlayerSprite_ctor(On.Celeste.PlayerSprite.orig_ctor orig, PlayerSprite self, PlayerSpriteMode mode)
     {
         orig(self, mode);
-
-        self.Animations[ELYTRA_ANIM] = new()
-        {
-            Frames = GFX.Game.GetAtlasSubtextures("characters/player_no_backpack/CommunalHelper/fly").ToArray(),
-            Delay = 10f,
-        };
     }
```
mayromr: This hook does nothing and should be removed.

src/States/Elytra.cs
```diff
@@ -466,9 +468,11 @@ private static void Mod_Player_UpdateSprite(On.Celeste.Player.orig_UpdateSprite
                 frame -= (int)(t * (FRAME_COUNT - STABLE_FRAME - 1));
             else
                 frame -= (int)(t * STABLE_FRAME);
-        }
-
-        self.Sprite.SetAnimationFrame(Calc.Clamp(frame, 0, FRAME_COUNT - 1));
+        }
+        if (self.Sprite.CurrentAnimationTotalFrames !< frame)
```
mayromr: I'll be real, I have no idea what does the `!<` do. is it the same as `>=` or just `<`. And is there a way to represent it in a more conventional.

Wait is the the `!` in `!<` just a null forgiving operator?

mayromr: Is this condition mandatory? Doesn't `Clamp` handle the case where `frame` is larger than `CurrentAnimationTotalFrames`?

mayromr (commented):
Hey, thanks for the contribution!

src/States/Elytra.cs
mayromr: I can't figure out how to change the comment lines :sweat_smile: I am talking about line 472.

AAA1459: > Is this condition mandatory? Doesn't `Clamp` handle the case where `frame` is larger than `CurrentAnimationTotalFrames`?

oh that is want to prevent animation from not existing in those frames by skin.

mayromr: Ah I see, shouldn't the condition be flipped then? because right now it will only call `SetAnimationFrame` if `frame` is larger then `CurrentAnimationTotalFrames`, or just  change the clamp values to `0` and `self.Sprite.CurrentAnimationTotalFrames - 1`.

AAA1459: > shouldn't the condition be flipped then? because right now it will only call `SetAnimationFrame` if `frame` is larger then `CurrentAnimationTotalFrames`

i guess "self.Sprite.CurrentAnimationTotalFrames !< frame" be like "!(self.Sprite.CurrentAnimationTotalFrames < frame)"... so what?

AAA1459: and sorry for replied now, I'm relatively busy

src/States/Elytra.cs
```diff
@@ -431,19 +431,26 @@ private static PlayerDeadBody Mod_Player_Die(On.Celeste.Player.orig_Die orig, Pl
...
+    }
+
+    private static void Mod_PlayerSprite_CreateFramesMetadata(On.Celeste.PlayerSprite.orig_CreateFramesMetadata orig, string sprite)
+    {
+        SpriteData data = GFX.SpriteBank.SpriteData[sprite];
+        if (!data.Sprite.Has(ELYTRA_ANIM))
+        {
+            string path = !string.IsNullOrEmpty(data.Sources[0].OverridePath) ? data.Sources[0].OverridePath : data.Sources[0].Path;
+            if (GFX.Game.HasAtlasSubtextures($"{path}CommunalHelper/fly"))
+                path = "characters/player_no_backpack/";
```
mayromr:
```suggestion
            if (!GFX.Game.HasAtlasSubtextures($"{path}CommunalHelper/fly"))
                path = "characters/player_no_backpack/";
```
Feel like you wanted to fallback to the `player_no_backpack`, but right now you always use that sprite.

src/States/Elytra.cs
```diff
@@ -466,9 +473,11 @@ private static void Mod_Player_UpdateSprite(On.Celeste.Player.orig_UpdateSprite
                 frame -= (int)(t * (FRAME_COUNT - STABLE_FRAME - 1));
             else
                 frame -= (int)(t * STABLE_FRAME);
-        }
-
-        self.Sprite.SetAnimationFrame(Calc.Clamp(frame, 0, FRAME_COUNT - 1));
+        }
+        if ((self.Sprite.CurrentAnimationTotalFrames - 1) >= frame)
```
mayromr:
```suggestion
        if (frame < self.Sprite.CurrentAnimationTotalFrames)

```

Tiny nit, even though I still prefeare limiting it via the `Clamp` without the `if`.

mayromr (commented):
Hey again, I feel like I am not qualified to review this anymore because I am not super familiar with the sprite system, maybe someone else could pick this up. But I do wonder if we sould iterate over sprite sources to search for metadata.

src/States/Elytra.cs
AAA1459: > Feel like you wanted to fallback to the `player_no_backpack`, but right now you always use that sprite.

Ah sorry for that, done now

mayromr:
Aight some food for thought:
1. What do we do if there are multiple SpriteData Sources? currently you only cack the first source but someone could have many skin mods installed maybe we should loop over all the SpriteDate sources like celeste does internally?
2. Currently if a mod doesn't have a custom elytra animation you inject the default `"characters/player_no_backpack/"` animation path, but the in `src/CommunalHelperGFX.cs/` `Mod_PlayerSprite_CreateFramesMetadat` we add the metadata to the sprite the mod defined. But if they don't define a custom elytra animation we will add the hair animation to the custom mode and then add the default elytra animation which will have no hair metadata.
3. Currently the game expects 9 diffrent frames of animation on the elyra flight, 2 of them for going up 1 of them for stable flight and 6 other for going down. should we un-hardcode it and say that  two third of the animation sprites are for the down animation the next one is for stable flight and the rest are for the up animation? and if do that, do we need to change the calculation for the current active frame?

I feel like a good way to approach this issue is to:
1. Combine the logic of loading the sprite into one `On.Celeste.PlayerSprite.CreateFramesMetadata` hook.
2. That hook will probe the currently loading sprite for an elytra animation if it found an elytra animation check if it has the correct hair frame metadata if it does do nothing, else inject the default hair frame metadata. If the animation isn't found inject the default elytra animation of the corresponding player sprite ( player / no_packpack / badeline ) and load the default metadata as well.

But take everything here with a grain of salt I very unfamiliar with the Monocle sprite system.

AAA1459:
> 1. What do we do if there are multiple SpriteData Sources? currently you only cack the first source but someone could have many skin mods installed maybe we should loop over all the SpriteDate sources like celeste does internally?
> 2. Currently if a mod doesn't have a custom elytra animation you inject the default `"characters/player_no_backpack/"` animation path, but the in `src/CommunalHelperGFX.cs/` `Mod_PlayerSprite_CreateFramesMetadat` we add the metadata to the sprite the mod defined. But if they don't define a custom elytra animation we will add the hair animation to the custom mode and then add the default elytra animation which will have no hair metadata.
> 3. Currently the game expects 9 diffrent frames of animation on the elyra flight, 2 of them for going up 1 of them for stable flight and 6 other for going down. should we un-hardcode it and say that  two third of the animation sprites are for the down animation the next one is for stable flight and the rest are for the up animation? and if do that, do we need to change the calculation for the current active frame?

1. Don't worry about if has multiple SpriteData Sources, Because it is impossible for two ID same in SpriteBank, the everest also just overwrite the that instead of adding more.

2. hair Metadate only base on fullpath of texture, and default texture already has it, so this is not a problem

3. I already did that in new commit, about making 2/3 frames for going down instead of No.0~5 frames etc

AAA1459:
> > 1. What do we do if there are multiple SpriteData Sources? currently you only cack the first source but someone could have many skin mods installed maybe we should loop over all the SpriteDate sources like celeste does internally?
> > 2. Currently if a mod doesn't have a custom elytra animation you inject the default `"characters/player_no_backpack/"` animation path, but the in `src/CommunalHelperGFX.cs/` `Mod_PlayerSprite_CreateFramesMetadat` we add the metadata to the sprite the mod defined. But if they don't define a custom elytra animation we will add the hair animation to the custom mode and then add the default elytra animation which will have no hair metadata.
> > 3. Currently the game expects 9 diffrent frames of animation on the elyra flight, 2 of them for going up 1 of them for stable flight and 6 other for going down. should we un-hardcode it and say that  two third of the animation sprites are for the down animation the next one is for stable flight and the rest are for the up animation? and if do that, do we need to change the calculation for the current active frame?
> 
> 1. Don't worry about if has multiple SpriteData Sources, Because it is impossible for two ID same in SpriteBank, the everest also just overwrite the that instead of adding more.
> 2. hair Metadate only base on fullpath of texture, and default texture already has it, so this is not a problem
> 3. I already did that in new commit, about making 2/3 frames for going down instead of No.0~5 frames etc

mayromr:
> 
> 1. Don't worry about if has multiple SpriteData Sources, Because it is impossible for two ID same in SpriteBank, the everest also just overwrite the that instead of adding more.
> 2. hair Metadate only base on fullpath of texture, and default texture already has it, so this is not a problem
> 3. I already did that in new commit, about making 2/3 frames for going down instead of No.0~5 frames etc

1. Not totally sure about that, I managed to load 2 SpriteData with 2 skin mods installed at the same time. Everest will overwrite base game sprites but if 2 modes try to overwrite the same sprite ( for example `player` ) I think it will let them.
2. Take a look at the most popular skin mod [niko](https://gamebanana.com/mods/251814), download it and load it with the code from this PR and then use the elytra. you will get the default player_no_backback sprite but without the hair.
3. Great! I haven't noticed that, but shouldn't the calculation be `FRAME_COUNT * (2f / 3f)`? because we have 6 frames of downward movement ( 0-5 )  1 stable frame ( 6 )  and 2 upward movement frames ( 7 - 8 ). This means that we took the stable frame from the upward 1/3 frames and not the downward  2/3 frames. So if we scale it up to 27 frames ( 3 times more frames ) for example, We would expect the downward movement frames to grow by 3 to 18, the stable frame should be the 19th frame ( frame in index 18 because we are counting from 0 ) and the rest of the frames are the upwards frames 27 - 18 - 1 = 8 (third of the frame minus the stable frame).

AAA1459:
> 2. Take a look at the most popular skin mod [niko](https://gamebanana.com/mods/251814), download it and load it with the code from this PR and then use the elytra. you will get the default player_no_backback sprite but without the hair.
> 3. but shouldn't the calculation be `FRAME_COUNT * (2f / 3f)`?

2. hmm indeed i can carefully think out that why, i'll fix it, just I can't run it because my VS2019 doesn't recognize part of the format
3. you should thinks: 
* FRAME_COUNT(9) are maximum No.8 from No.0 counting.
* STABLE_FRAME(6) are No.6 from No.0 counting. if like FRAME_COUNT from No.1 counting so it are 7.
so you sure STABLE_FRAME need get 6 by (9 * (2f/3f)) instead (9 * (7f/9f))-1?

mayromr:
> 3. you should thinks:
> 
> * FRAME_COUNT(9) are maximum No.8 from No.0 counting.
> * STABLE_FRAME(6) are No.6 from No.0 counting. if like FRAME_COUNT from No.1 counting so it are 7.
>   so you sure STABLE_FRAME need get 6 by (9 * (2f/3f)) instead (9 * (7f/9f))-1?

I am almost certain about it, `(FRAME_COUNT * (7f/9f))` finds the first index of the up animation ( frame 7 ) and then you remove 1 to find the `STABLE_FRAME` but this isn't always true, for example if we want to triple the animation frames to 27 we would expect `STABLE_FRAME` to triple as well to 18 ( 6 * 3 ). But with the `(27 * (7f/9f)) - 1` we would get 20 which is off by 2, the error comes from the fact that each element in the array became 3 elements in the new array so `STABLE_FRAME` ( frame 6 ) became frames 18, 19 and 20,  so we can no longer go backwards one step and expect to find the **first** cell that represents cell 6  ( as there are now 3 of them now )

AAA1459:
> I am almost certain about it, (FRAME_COUNT * (7f/9f)) finds the first index of the up animation ( frame 7 ) and then you remove 1 to find the STABLE_FRAME but this isn't always true, for example if we want to triple the animation frames to 27

why are you obsessed with 9 * 3 == 27 ?   you should try to thinks:
9 * 1.5f == 13
13 / 9f * 7f == 10______STABLE_FRAME+1______((13 / 3f * 2f) == 8) + 1 == 9
13 - 10 == (2 * 1.5f)______going_up______13 - 9 == (2 * 1.5f) + 1
13 - 3 - 1 == (6 * 1.5f)______going_down______13 - 4 - 1 == (6 * 1.5f) - 1
So whose value do you think is more true for STABLE_FRAME about (x*(7/9)-1) and (x*(2/3)) ?

mayromr:
There is no "correct" answer here, again because there is *exactly one*  `STABLE_FRAME` we will almost always have a mismatch with the scaled number of elements in the UP or DOWN images, this is because if we scale `STABLE_FRAME` ( & `FRAME_COUNT` ) up by some `X` but only used one of those possible values so we are left with `X - 1` additional elements that need to fit neatly in UP and DOWN, but if `X - 1` isn't divisible by 3 we will never be able keep the original ratio ( we try to split 2/3 DOWN 1/3 UP ). 
The big difference between `FRAME_COUNT * (2 / 3)` and  `FRAME_COUNT * (7 / 9) - 1` is that the former is adding the additional frames to the UP part and the latter is adding them to the DOWN part that's it. Now my point is that it's easier for the modder to find the rounded down 2/3 point then the rounded down 7/9 - 1 point.
Regarding your calculation it's a bit hard to read but I think you forgot to take the minus 1 into account on your side, and on top of that I wanted to use 27 because it is divisible by 9 ( & 3) to keep thinks simple, that's it, you can use any other number you want as long as `X - 1` isn't divisible by 3 there will be left with an "error".  
TLDR: idrc which one you chose, tbh I feel like 99% of all skin mods will use 9 frames of animation and be done with it.

AAA1459:
> tbh I feel like 99% of all skin mods will use 9 frames of animation and be done with it.

I agree with that, and it's too much trouble and redundant to discuss, so now let's ignore it ()

vivianlonging:
This has been partially resolved within the recent dev push. Not closed.
While this seems less "stable" than the solution provided, this is a clean enough system to manage the manual addition of the fly animations, as there is no Everest-supported Appending of Animations to a pre-existing spriteData within Sprites.xml

With respect to the fly animations, I think that I'll have to support two different formats now, however I'm okay with this as a more permanent solution.

My ideas for managing the flight animations is redefining the set, possibly with a more complex but appropriate solution. The simplest case would be:
- flyUp
- flyTransitionUp
- flyStraight
- flyTransitionDown
- flyDown

but this can be elaborated on. To that end, personally I think we should try to make it as simple as possible for non-coders to implement flying animations, so doing something like fly00-08 feels, janky at best and restrictive at worst. That being said, I need to update for SSC *today* so we're going with the 9-frame solution temporarily.

Graphics/Atlases/Gameplay/characters/player_no_backpack/CommunalHelper/reskin_guide.txt
catapillie, line 1: Consider making this file a Markdown one (`.md`) so that it displays nicely on GitHub when navigating to this directory. Also, if you haven't already, you could edit out Wiki page.

Graphics/Atlases/Gameplay/characters/reskin_guide.txt
catapillie, line 1: Is this a duplicate?

catapillie (commented):
I don't have the time to look at the code for now, but I will in the near future. Thanks for adding the new sprites and tweaking the previous ones.

Graphics/Atlases/Gameplay/characters/reskin_guide.txt
AAA1459: It's a duplicate, I don't remember why I place it there...

(edited: maybe i was thinks the user will't 100% open player_no_backpack that one and see it)

Graphics/Atlases/Gameplay/characters/player_no_backpack/CommunalHelper/reskin_guide.txt
AAA1459: Ah I never thought about that, thanks you remind me I'll try it now

catapillie (commented):
Thanks for removing the files

PR #172, opened 2023-11-23
Add DeployElytraBinding to bird tutorial command

### LINKED ISSUE #154: Elytra keybind keyword for tutorials
We need a keyword for displaying elytra keybinds in tutorial birds and similar.

- [x] Is your request a change to an existing feature?
  - Elytra / (Everest bird tutorial parsing event)

mayromr:
closes #154, the new keyword is `CommunalHelperDeployElytraBinding`.

### DIFF
--- a/src/CommunalHelperModule.cs
+++ b/src/CommunalHelperModule.cs
@@ -337,15 +337,18 @@ public class CommunalHelperModule : EverestModule
     private object CustomBirdTutorial_OnParseCommand(string command)
     {
         // Thank you maddie.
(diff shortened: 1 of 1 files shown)


vivianlonging:
ngl you shouldn't have bothered with this I should have been more active in the Issues tab.
You can already do this with mod:CommunalHelper/DeployElytra (i'm fairly sure)
I'll make a test of that though just to be sure, maybe Grab Modes break it

mayromr:
Ah I see, ya `mod:CommunalHelper/DeployElytra` works. I guess we should close #154 then?

PR #175, opened 2023-12-18
Fix linking against core and bump Dotnet and Everest version

mayromr:
Linking was broken for me for me even with `legacyRef` so I used the opportunity to bump Dotnet and Everest version

### DIFF
--- a/everest.yaml
+++ b/everest.yaml
@@ -1,6 +1,6 @@
 - Name: CommunalHelper
   Version: 0.0.0-dev
-  DLL: src/bin/Debug/net452/CommunalHelper.dll
+  DLL: src/bin/Debug/net7.0/CommunalHelper.dll
   Dependencies:
     - Name: Everest
-      Version: 1.3876.0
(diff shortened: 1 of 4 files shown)


everest.yaml
```diff
@@ -1,6 +1,6 @@
 - Name: CommunalHelper
   Version: 0.0.0-dev
-  DLL: src/bin/Debug/net452/CommunalHelper.dll
+  DLL: src/bin/Debug/net7.0/CommunalHelper.dll
   Dependencies:
     - Name: Everest
```
l-Luna: Might need to change dependency name to `EverestCore`?

l-Luna (approved):

everest.yaml
mayromr: Right my bad, just [looked at](https://github.com/EverestAPI/CelesteModTemplate/blob/main/CelesteMod/everest-core.yaml) I'll update it and test that it works

PR #178, opened 2024-02-16
Add "Reusable" property to "TimedTriggerSpikes"

Saplonily:
This pr added "Reusable" property to the entity "TimedTriggerSpikes".
The "Reusable" property allows users to retract the timed spikes after it's triggered, and the retract delay equals to the "Reusable Timer", if the 'Reusable' property isn't enabled, then the spikes work as normal timed spikes.
~~Sorry for my poor julia that I didn't add support for Ahorn~~

### DIFF
--- a/Loenn/entities/misc/timed_trigger_spikes.lua
+++ b/Loenn/entities/misc/timed_trigger_spikes.lua
@@ -15,6 +15,8 @@ for _, dir in ipairs(directions) do
         placement.data.WaitForPlayer = false
         placement.data.Grouped = false
         placement.data.Rainbow = false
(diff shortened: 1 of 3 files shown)


vivianlonging:
Personally, I think we can reduce the property window of the TimedTriggerSpikes simply by removing the Reusable attribute and making ReusableTimer "fail" at values less than 0.
I can make Ahorn plugin support for it as needed before the Merge is finalized

Saplonily:
I think we can keep the selection box? cause that sounds more intuitive for mappers to use, besides, there aren't really that much parameters to set in this spike.

PR #181, opened 2024-03-12
ReverseHelper/ReversedDreamBlock compatibility 

wuke32767:
You can also choose not to add an optional dependency. There're already compatibility stuff before ReverseHelper adding interop.
I'm about to remove them after CommunalHelper updated.


catapillie (approved):
Thanks! What do you mean about the optional dependency? - I didn't quite understand. Is there something you still need to commit?

wuke32767:
Sorry, just ignore that. I thought optional dependency is necessary for modinterop.
Just ignore the initial message. I also forgot why i said that.
there's nothing else to do.

catapillie:
Okay, thanks!

PR #184, opened 2024-04-27
SJ laser emitter port

swoolcock:
Port of the `LaserEmitter` entity from SJ, namely those used in Lethal Laser Laboratory and In Filtration.  Currently only the "rounded" style is available by default, but the sprites can be customised.

Trying to get a useable version of this in dev/beta fairly soon since it's blocking a mapper.

The `Paintbrush` entity from paint uses similar code, however it's more effort to port due to all the custom rendering and cassette block handling, so that's a future task.


catapillie (approved):
Everything looks good to me.

PR #186, opened 2024-04-29
Fix cassette entity rendering in loenn + update lang file description

aonkeeper4:


catapillie:
Perfect, thank you so much!

PR #188, opened 2024-05-17
Implement camera culling for Connected Solids

JaThePlayer:
Simple broad-phase camera checks for better perf in large rooms. Better performance for large solids could be achieved by checking which parts of the solids are visible and only rendering those, but this will do for now.

Also fixes a bug where zip mover borders wouldn't be rendered if its path was off-screen, even if the block itself was on-screen


catapillie (approved):
Thanks!

PR #190, opened 2024-06-26
fix dream move block line color overwriting global dream line color

karl2883:
Previously, the static dream block attribute activeLineColor was permanently overriden with the "wobbleLineColor" attribute (the current color of a dream move block, defaults to white) by the CustomDreamBlock class. To fix this, the wobbleLineColor attribute was moved out of CustomDreamBlock into DreamMoveBlock and the static activeLineColor gets reverted to its old value after rendering a dream move block. This will fix the issue of forced white outline colors with CustomDreamBlocks on screen (https://discord.com/channels/835962669453410354/1244848147327619154).


catapillie (approved):
Untested but I'll merge given the circumstances

catapillie (approved):

PR #194, opened 2024-07-27
Add paintbrushes from SJ

swoolcock:
The FMOD bank contains all events from paint, so I've added event strings for all of them.  It's possible someone will want the other paint entities ported in future anyway.

* Sound assets ported with permission from BeanJammin.
* Art assets ported with permission from mosscairn, however moss specifically asked for their name to be included in the file path.


catapillie (approved):
LGTM

PR #198, opened 2024-08-16
add Two Dash option to dream refill

vitellaryjr:
two-dash dream refills will give two dream tunnel dashes for the player to use consecutively. there's currently no real indicator for how many dream dashes you have left, but i'm not sure how necessary it is to have one since you can never get more than 2 with this current code. if there's any ideas for what indicator could work i'm willing to hear it out

this involved changing the HasDreamTunnelDash boolean into a DreamTunnelDashCount integer, and adding a DreamTunnelDoubleDash DashState. i'm hoping i handled everything involving those changes well enough; the mod exports were not changed functionally (the previous HasDreamTunnelDash function still exists, and just returns whether you have more than 0 dream dashes available), so hopefully this shouldn't break any compatibilities with other mods, if such things exist

i don't actually need this merged any time soon, i'm not even sure if i'm going to be using it myself. i just thought it would be cool to have and at least one other person also expressed that so i made it


vitellaryjr:
one thing that i have noticed with this change is that it doesn't play well with savestates (with dream dash state being kept between reloads); unsure if this is something that needs to be changed on my end, or if it requires a change to speedrun tool

vitellaryjr:
figured out how to add savestate compatibility within communal helper, so this feature should now be ready to go

this commit also adds a small indicator for whether you have multiple dream dashes, where your dream dash trail will be created more often. didn't test with values above 2 because i don't think it's possible to have more than 2, but it should provide a noticeable difference at higher values too

PR #210, opened 2024-10-18
Fix move block debris lag

aonkeeper4:
Adds a cap to debris vertical speed, reducing lag when falling.

### DIFF
--- a/src/Entities/Misc/MoveBlockDebris.cs
+++ b/src/Entities/Misc/MoveBlockDebris.cs
@@ -73,18 +73,22 @@ public class MoveBlockDebris : Actor
     {
         base.Update();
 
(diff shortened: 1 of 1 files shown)


bigkahuna443:
I tested it out and it looks a bit visually funky compared to vanilla move blocks (especially if both are used together). I'd prefer it if visuals were unaffected while debris is on-screen

aonkeeper4:
yeah, makes sense. i think there was a reason i didn’t just have them stop
completely or something if offscreen but idr
will take another look 👍

aonkeeper4:
@bigkahuna443 changed it to check if in level bounds instead of capping speed so now there should be no visual inconsistencies :+1:

PR #213, opened 2024-11-01
When the curvebubble moving vertically or horizontally, avoid it make the camera twitch or won't hitting certain spikes

AAA1459:


src/Utils/BakedCurve.cs
```diff
@@ -158,7 +157,12 @@ public void GetAll(float t, out Vector2 point, out Vector2 derivative)
         {
             point = GetCurvePoint(points[i], points[i + 1], points[i + 2], percent);
             derivative = GetCurveDerivative(points[i], points[i + 1], points[i + 2], percent);
-        }
+        }
+        // Correct very small deviations; When the curvebubble moving vertically or horizontally, avoid it make the camera twitch or won't hitting certain spikes.
+        if (Math.Abs(derivative.X) < 0.004f)
+            derivative.X = 0f;
+        if (Math.Abs(derivative.Y) < 0.004f)
+            derivative.Y = 0f;
```
bigkahuna443, line 165: I think this change should be made in the Booster classes themselves and not this utility function. I'm also a bit unsure about how this might effect current maps (although the change seems reasonable)

catapillie:
I more or less agree with @bigkahuna443 on this. Should be changed to affect only boosters that have this problem, and not change the implementation of baked curves (which I suppose aimed to be mathematically correct).

PR #215, opened 2024-11-07
Use Everest's state-registry API instead of manually creating them

psyGamer:
This reduces potential conflict with other mods which register states in the legacy way, and also allows dropping the CelesteTAS dependency to provide state information to the Info HUD.
I'm not sure which Everest version introduced that API, so just to be safe, depend on the latest stable.

### DIFF
--- a/src/CommunalHelper.csproj
+++ b/src/CommunalHelper.csproj
@@ -32,9 +32,6 @@
 		<Reference Include="$(CelestePrefix)\MMHOOK_Celeste.dll">
 			<Private>false</Private>
 		</Reference>
-		<Reference Include="$(CelestePrefix)\Celeste.dll">
-			<Private>false</Private>
-		</Reference>
 		<Reference Include="$(CelestePrefix)\FNA.dll">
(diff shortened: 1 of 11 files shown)


src/CommunalHelper.csproj
```diff
@@ -32,9 +32,6 @@
 		<Reference Include="$(CelestePrefix)\MMHOOK_Celeste.dll">
 			<Private>false</Private>
 		</Reference>
-		<Reference Include="$(CelestePrefix)\Celeste.dll">
```
bigkahuna443: Not sure why this was removed? Would prefer to have any build changes done in separate PRs

psyGamer: Because `Celeste.dll` is already included through the publicizer.
I dont't remember if it actually caused issue for me, but it's better to not include Celeste twice.

bigkahuna443: Gotcha. I'll undo it in this PR and include it in a separate commit.

PR #219, opened 2025-01-02
Allow Move Block Groups to affect all custom move block types (no longer breaks Move Block Redirects)

aonkeeper4:
This is essentially PR #206 with the issues regarding Move Block Redirects fixed and a couple minor tweaks I didn't get to put in the first PR.
Allows Move Block Groups to affect all custom move block types (Connected Move Blocks, Equation Move Blocks, Cassette Move Blocks, Dream Move Blocks, and Move Swap Blocks) via a new GroupableMoveBlock component. Adds indicators on these other move block types to show the group they are part of, in the style of the existing one on Connected Move Blocks.


bigkahuna443:
This part of the code has always been terrifying to me lol. Seems good to me from what I can tell and fixes the issues from the last PR!

PR #222, opened 2025-01-05
Add more options to Dream Tunnel Dash and Elytra

### LINKED ISSUE #137: DreamTunnelDash does not let you buffer a hyper
When Madeline sits upon a dream block, she can input a down diagonal dash and buffer a jump during the freeze frames which results in a hyper instead of going through the dream block. When Madeline sits upon normal ground after grabbing a dream dash crystal, buffering a jump after a down diagonal dash still causes her to dream tunnel through the ground.

- [x] Is your request a change to an existing feature?
  - DreamTunnelDash
- [ ] Is this something that will be used for an upcoming collab/contest?

**Additional details:**
This is mostly for consistency with the base game and can clear up some confusion I had with The Tower XVI from SJ2021. I can probably make the change myself if I knew where to look but alas I'm not sure where to even start for this.

aonkeeper4:
Adds more customizability to the Dream Tunnel Dash and Elytra, configurable by the (preexisting) Configure Elytra and (new) Configure Dream Tunnel Dash triggers.
These new options are explained in the tooltips for the trigger attributes, but in brief:
- Dream Tunnel Dash now supports many of the options provided in Frost Helper's Custom Dream Blocks and Pandora's Box's Dream Dash Controller, as well as an option to allow cancelling the dash before entering solids (replicating being able to instant hyper on dream blocks).
- Elytra now supports disabling the vertical boost the player gets when deploying it facing the opposite direction to their speed.


bigkahuna443:
I'm not sure about the positioning of the dash cancel hook either, but it seems to work fine and I wasn't running into any odd behavior with my testing. Thanks for the PR! 

Also I believe this closes #137

src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
```diff
@@ -186,12 +239,27 @@ private static void Player_CreateTrail(On.Celeste.Player.orig_CreateTrail orig,
...
+        /*
+         * start the player dream tunnel dash later if needed to allow cancelling it
+         * this replicates vanilla behavior of being able to instant hyper for example on dream blocks
+         * inserts a call to StartDreamTunnelDashAttacking after the call to CallDashEvents if the current dream dash config allows dash cancels
+         * is this the best place to put this logic?
+         */
+        ILLabel afterStartDashAttack = cursor.DefineLabel();
+        cursor.GotoNext(MoveType.After, instr => instr.MatchCallvirt<Player>("CallDashEvents"));
+        cursor.EmitDelegate(() => CommunalHelperModule.Session.CurrentDreamTunnelDashConfiguration.AllowDashCancels);
+        cursor.Emit(OpCodes.Brfalse, afterStartDashAttack);
```
wuke32767, line 253: ~~brtrue imo~~

wuke32767: ~~wait, whats the point of this part~~
alright, but *this* StartDreamTunnelDashAttacking should not check for the solid

aonkeeper4: hang on, why should this not check for the solid? to my understanding the fast-moving solid check is for cases when the block would move out of the way of the player's dash before they could collide with it due to update order - is there any reason why we shouldn't be checking for that here? i agree that speed and dashdir shouldn't be overridden in this case though

wuke32767: the fast-moving check is for the first frame, player has no speed, so player can't collide with any solid.

here it's the second frame, player has speed. do the check will lose the speed (see below). *(alright, just don't clear the speed)* 

https://github.com/CommunalHelper/CommunalHelper/blob/aa8b1f383891ba8213813404cab3ac92ec1929f7/src/DashStates/DreamTunnelDash/DreamTunnelDash.cs#L190

PR #225, opened 2025-01-25
Change state names

megamaz:
Changed state names from `Elytra` to `StElytra` and `DreamTunnelDash` to `StDreamTunnelDash` to make it clearer that they are states in CelesteTAS.

I do not mod Celeste, so I don't know if there are more references to this that need to be updated to make this fully function. The goal is just to make it easier to differentiate that `Elytra` is a state.

### DIFF
--- a/src/States/St.cs
+++ b/src/States/St.cs
@@ -24,7 +24,7 @@ public static class St
 
     private static void RegisterPlayerStates(Player player)
     {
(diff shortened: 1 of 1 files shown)


psyGamer:
The "St" shouldn't be part of the name, as seen with vanilla: https://github.com/EverestAPI/Everest/blob/152f3e73b68902d0c872f80799234c71a1d5e81e/Celeste.Mod.mm/Patches/Player.cs#L63-L88
But CelesteTAS could add that prefix.

bigkahuna443:
Yeah I'm gonna trust psyGamer's expertise on this one 👍

PR #226, opened 2025-01-25
Dream Tunnel Dash fixes + new "Redirect Consumes Normal Dash" option

aonkeeper4:
Fixes a couple of bugs with the new Dream Tunnel Dash features, namely wrong fast-moving solids check logic with the "Allow Dash Cancels" option and inconsistencies with dream grabbing with the "Allow Redirects" option. Also adds the option to consume normal dashes instead of dream tunnel dashes when redirecting.
i have no clue why the diff is so gigantic i swear i only changed like 20 lines total weheline


aonkeeper4:
evil line ending inconsistencies have been fixed :glumbsup:

PR #227, opened 2025-03-04
Add option to remove debris to all move block types

aonkeeper4:
debris is laggy. simple solution imo


catapillie (approved):
Seems all good to me. Thanks for this!

PR #230, opened 2025-03-20
Kosei/new attributes

koseidiamond:
This PR sets a flag to true while a textbox from the Hints controller is being read.

It also adds a flag that is toggled when a track swap box is dashed, and fixes the "bouncy" attribute not doing anything (it was called "bounce" internally).

### DIFF
--- a/Loenn/entities/misc/hint_controller.lua
+++ b/Loenn/entities/misc/hint_controller.lua
@@ -13,8 +13,9 @@ hintController.placements = {
             singleUses = "",
             selectorCounter = "",
             selectNextHint = false,
-        },
-    },
+			flagWhileReading = ""
+        }
+    }
 }
 
 return hintController
--- a/Loenn/entities/misc/hint_controller.lua
+++ b/Loenn/entities/misc/hint_controller.lua
@@ -14,6 +14,7 @@ hintController.placements = {
             selectorCounter = "",
             selectNextHint = false,
 			flagWhileReading = ""
+            flagWhileReading = "",
         }
     }
 }
--- a/Loenn/entities/station_blocks/track_switch_box.lua
+++ b/Loenn/entities/station_blocks/track_switch_box.lua
@@ -15,7 +15,7 @@ trackSwitchBox.placements = {
             floaty = true,
             bouncy = true,
             reverse = false,
-			switchFlag = ""
+            switchFlag = "",
         }
     },
     {
--- a/Loenn/entities/station_blocks/track_switch_box.lua
+++ b/Loenn/entities/station_blocks/track_switch_box.lua
@@ -14,7 +14,8 @@ trackSwitchBox.placements = {
             globalSwitch = false,
             floaty = true,
             bouncy = true,
-            reverse = false
+            reverse = false,
+			switchFlag = ""
         }
     },
     {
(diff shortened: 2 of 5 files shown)


earthwise01:
regarding the fix for the "bouncy" attribute not doing anythingg  i think it might be worth still checking the old "bounce" attribute with something like `data.Bool("bounce", data.Bool("bouncy", true))` in order to not break existing maps

from what i can remember "bounce" was what it was originally named? but it seems like when the lönn plugin got made "bouncy" was used by mistake without adjusting the code to match

earthwise01:
actually thinking about this more that might still break existing maps? since the "bouncy" checkbox is enabled by default in lönn despite not working, this would still make it so track switch boxes which would've been stationary before will suddenly start bouncing

i think maybe leaving the code as `data.Bool("bounce")` and changing the *lönn plugin* to use "bounce" instead should mean any preexisting entities stay as they are but newly placed ones are able to bounce
(in that case it might also be worth adding a lang entry to label the existing "bouncy" attribute broken? eg `...attributes.name.bouncy=Bouncy (broken)`)

koseidiamond:
No, it's the opposite case - by default, all of them used to be bouncy. The default value was (and still is) true both in the plugin and in the code. This change would mean that old maps that have the bouncy attr unset would stop bouncing, matching the intended behavior and not breaking the default. But maybe you're right and for legacy it's better to replace the attribute with a new one in lönn.

earthwise01:
huh that doesn't seem to be the case for me though? if i try placing a new track switch box down, while it has the 'bouncy' checkbox on by default, in-game it always stays in place when i dash into it no matter what i set it to

from what i can tell yes the code sets the `canBounce` field to `true`, but then always immediately overwrites it with the value returned by `data.Bool("bounce")` in the constructor (which will always be `false` if there isn't a "bounce" attribute in the entity data)

if `data.Bool("bouncy")` is used instead then that would mean any existing track switch boxes which didn't have their bouncy attribute touched will suddenly start bouncing even if they weren't doing so before

Loenn/entities/misc/hint_controller.lua
```diff
@@ -13,8 +13,9 @@ hintController.placements = {
             singleUses = "",
             selectorCounter = "",
             selectNextHint = false,
-        },
-    },
+			flagWhileReading = ""
```
catapillie, line 16:
```suggestion
			flagWhileReading = ""
            flagWhileReading = "",
```

Loenn/entities/station_blocks/track_switch_box.lua
```diff
@@ -14,7 +14,8 @@ trackSwitchBox.placements = {
             globalSwitch = false,
             floaty = true,
             bouncy = true,
-            reverse = false
+            reverse = false,
+			switchFlag = ""
```
catapillie:
```suggestion
            switchFlag = "",
```

Loenn/entities/station_blocks/track_switch_box.lua
```diff
@@ -23,7 +24,8 @@ trackSwitchBox.placements = {
             globalSwitch = true,
             floaty = true,
             bouncy = true,
-            reverse = false
+            reverse = false,
+			switchFlag = ""
```
catapillie:
```suggestion
           switchFlag = "",
```

catapillie (approved):
Thanks for the implementation & the fixes. Don't mind me just fixing some indentation

PR #231, opened 2025-03-21
cloudscape zoom out support (fixed)

earthwise01:
see #216.

re-adds cloudscape zoom out support with a few fixes, namely vanilla zoom should work correctly again and zoom behavior "adjust" now zooms out from the center of the visible area instead of the center of the cloudscape (which looked weird).


catapillie (approved):

catapillie:
Thanks a lot. I haven't tested it, though

PR #232, opened 2025-03-25
Add Dream Theo Crystal

aonkeeper4:
Adds a Dream Theo Crystal that works similar to Dream Jellyfish, except being a Theo Crystal.
To facilitate this, this PR also:
- Refactors the dream sprite rendering used by Dream Jellyfish into the new DreamSprite component:
  - This component takes a Sprite holding solid white textures and renders it (via DreamSpriteRenderer, which replaces DreamJellyfishRenderer) as a mask through which dream block particles are visible, with an outline.
  - Dream Jellyfish mask textures are no longer gigantic as DreamSpriteRenderer uses the stencil buffer instead of BlendState shenanigans.
- Refactors all "dream holdable" behavior into the new DreamHoldable component:
  - This includes the dream dash collider, collider inversion logic and collider "refilling" logic.
  
Use of DynamicData in Dream Jellyfish has also been removed.


catapillie (approved):
Looks really good, thanks for the refactor.
Have to test if inverted dream jelly fishes still have corrected offsets.

catapillie:
Works super well! thanks!
For some feedback, I think the dream theo crystal sprite could do with a bit less white outlines. I can barely see the dream star particles... Perhaps make some of the inner outlines transparent, like 50% alpha.

PR #233, opened 2025-04-02
Add 'Slow' option for Dash Zip Movers + Double On. hook fix

SunsetQuasar:
Adds a 'Slow' field for Dash Zip Movers that make them 1.75x slower. Also fixes a double On. hook from the Dream Theo implementation

### DIFF
--- a/Loenn/entities/strawberry_jam/dash_zip_mover.lua
+++ b/Loenn/entities/strawberry_jam/dash_zip_mover.lua
@@ -39,6 +39,7 @@ dashZipMover.placements = {
         ropeLightColor = "329415",
         ropeShadowColor = "003622",
         soundEvent = "event:/CommunalHelperEvents/game/strawberryJam/game/dash_zip_mover/zip_mover",
+        slow = false,
     }
 }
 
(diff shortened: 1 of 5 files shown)


catapillie (approved):
LGTM. I wonder if the `CubeIn` easing function makes the liftspeed match. I don't think so, but it'd be cool if a perfect match were your goal.

If we name our fast and slow factors respectively $f_1$ and $f_2$ then the zip mover's position is, for each factor:

$$x_1(t) = \left(f_1t\right)^2$$
$$x_2(t) = \left(f_2t\right)^\alpha$$

We want the derivatives to match when these functions reach 1, which gives

$$\frac{\text{d}x_1}{\text{dt}}\left(f_1\right)=\frac{\text{d}x_2}{\text{dt}}\left(f_2\right)$$

and you find

$$\alpha = \frac{2f_2}{f_1}$$

So, with $f_1=1.0$ and $f_2=1.75$, you should have $\alpha = 3.5$, not $3$ (`CubeIn`).
I don't mind it being 3 though. But you could make the zip mover time parameter perhaps

Thanks for the double ON hook fix!

SunsetQuasar:
i didn't know there was any math behing that LMAO i just changed it to CubeIn arbitrarily bc it felt appropriate. is there a way to create such an easer? if so, i think changing it to that would be nice

catapillie:
Yeah. AFAIK an Easer is just a function from [0, 1] to [0, 1]. Passing a delegate should work.
See: https://github.com/shortgecko/MonocleEngineMirror/blob/master/Util/Ease.cs

SunsetQuasar:
figured it out. done
made a custom easer maker in Util

catapillie (approved):
Thank you!

PR #234, opened 2025-04-06
Make Melvins reskinnable and add `MelvinTargetable` component so they can target non-player entities

aonkeeper4:
Adds options to make Melvins reskinnable and to change their particle colors, and adds a new `MelvinTargetable` component (along with ModInterop) to allow Melvins to target non-player entities.
Also fixes a bug with the LylyraHelper interop where the Loop Block particle color getter was assigned to Dream Falling Blocks instead.


wuke32767:
what do you think of adding priority options to the component

aonkeeper4:
ooh thats not a bad idea
ill have a look at that

src/Imports/LylyraHelper.cs
```diff
@@ -60,7 +60,7 @@ public static void Load()
                 }
             }
         };
-        RegisterSlicerActionSet(typeof(DreamFallingBlock), loopBlockDict);
+        RegisterSlicerActionSet(typeof(LoopBlock), loopBlockDict);
```
catapillie, line 63: thank you for this

src/Entities/Misc/Melvin.cs
```diff
@@ -765,55 +872,60 @@ public override void Update()
...
+                        crushDir = Vector2.UnitY;
+                        dir = ArrowDir.Down;
+                        toTargetRect = new Rectangle(x1, (int) (Y + Height), x2 - x1, (int) (target.Top - Y - Height));
+                    }
                 }
-            }
-            if (detectedPlayer && IsPlayerSeen(toPlayerRect, dir))
-            {
+                
+                if (!detectedTarget || !IsPlayerSeen(toTargetRect, dir))
```
catapillie, line 924: might want to rename the `IsPlayerSeen` auxiliary method

catapillie (approved):
Thanks a lot this is a very nice implementation

PR #235, opened 2025-04-10
Add option to allow screen transitions while dream tunnel dashing

aonkeeper4:
Adds a new option to the Configure Dream Tunnel Dash trigger to allow screen transitions in the dream tunnel dash state.
Also modifies the state check utility used in the dream tunnel dash hooks to allow checking for any state, as this is necessary for some hooks that are required for this feature and reusing stuff is cool.

### DIFF
--- a/src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
+++ b/src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
@@ -73,6 +73,7 @@ public static class DreamTunnelDash
         public float CustomSpeed;
         public bool AllowDashCancels;
         public bool RedirectConsumesNormalDash;
+        public bool AllowTransitions;
     }
 
     public static readonly DreamTunnelDashConfiguration DefaultDreamTunnelDashConfiguration = new()
@@ -85,6 +86,7 @@ public static class DreamTunnelDash
         CustomSpeed = 0,
         AllowDashCancels = false,
         RedirectConsumesNormalDash = false,
+        AllowTransitions = false,
     };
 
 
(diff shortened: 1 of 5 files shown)


src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
```diff
@@ -360,33 +368,100 @@ private static void Player_IsRiding_JumpThru(ILContext il)
...
+    
+    private static void NaiveMoveTowardsY(Player player, float targetY, float maxAmount, Collision _)
+    {
+        float toY = Calc.Approach(player.ExactPosition.Y, targetY, maxAmount);
+        player.movementCounter.Y += (float) ((double) toY - player.Position.Y - player.movementCounter.Y);
+        int y = (int) Math.Round(player.movementCounter.Y);
+        player.Position.Y += y;
+        player.movementCounter.Y -= y;
+    }
+
```
catapillie: Were you not able to use `NaiveMove` from the game's code? It's probably preferable

catapillie (commented):
Code looks okay, however I'm not going to act like I can tell if it's correct or not without some testing and ILSpy-ing. I'm a bit worried by the `NaiveMove` copy thing, what was the motivation behind it?
I probably need some time to test. I have lots of IRL stuff coming soon and might not be able to work on this for a while. In that case you should probably work with a debug build.

src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
aonkeeper4: yea thats a good shout, idr the motivation for this honestly

aonkeeper4: i think i thought it was weird using NaiveMove when i only needed to change one component

catapillie (approved):

src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
catapillie: Huh okay. I thought there was a `NaiveMoveH` and `NaiveMoveV` or something. Must have misremembered

catapillie:
It looks okay so I'll merge it.

PR #238, opened 2025-04-14
Added "No Hole" option to Loop Blocks

Spo0o0ky:
Updated the loenn plugin for Loop Blocks to include a boolean option that toggles whether a hole is rendered in the centre of the loop blocks, creating the appearance of one solid block.

image


CPCTC:
I'm not a maintainer here but minor nitpick: It might be better to call the option "With Hole", or maybe "Filled", instead of "No Hole", to avoid a double negative

Spo0o0ky:
I'm open for that to be changed, although I think "No hole" makes more sense as having a hole is the default which people would expect.

catapillie (changes requested):
Thanks a lot, looks all good. I'd like the Lönn plugin to reflect to `noHole` option though

catapillie (approved):

PR #239, opened 2025-04-27
zoom out support for Shape3Ds and DreamSprites

earthwise01:
adds zoom out support to Shape3D components (used by Aero Blocks, Shapeshifters, and Elytra Rings) and DreamSprite components (used by Dream Jellyfish and Dream Theo Crystals)


catapillie (approved):
I'm surprised so little was needed to make it compatible with zoom-out. Thanks!

PR #242, opened 2025-05-11
Implement Dream Tunnel Blockers

SnipUndercover:
Allows the mapper to prevent entering a dream tunnel/dream dash when colliding with solids/dream blocks covered by this entity.

> [!NOTE]
> If the player is already in the dream dash/dream tunnel state, this entity has no effect.

### DIFF
--- a/src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
+++ b/src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
@@ -1,9 +1,11 @@
 ﻿using Celeste.Mod.CommunalHelper.Components;
+using Celeste.Mod.CommunalHelper.Entities;
 using Celeste.Mod.CommunalHelper.States;
 using Mono.Cecil.Cil;
 using MonoMod.Cil;
 using MonoMod.RuntimeDetour;
 using MonoMod.Utils;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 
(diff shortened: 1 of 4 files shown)


src/DashStates/DreamTunnelDash/DreamTunnelDash.cs
```diff
@@ -740,6 +755,37 @@ private static bool DreamTunnelDashCheck(this Player player, Vector2 dir)
...
+                blockDreamTunnelDashes = true;
+            if (e is DreamTunnelBlocker { BlockDreamDashes: true })
+                blockDreamDashes = true;
+
+            if (blockDreamTunnelDashes && blockDreamDashes)
+                break;
+        }
+
+        return (blockDreamTunnelDashes, blockDreamDashes);
+    }
```
catapillie, line 775: Is this used at all?

catapillie (commented):
Looking okay for the most part. Though there are two things I'm confused by
1. there's an unused static method
2. where is the static dream tunnel blocker list actually populated? Is it the call to CollideAll that does it?
Apologies if these questions sound unusual, but I have actually not touched Celeste for a couple years.

SnipUndercover:
1. I'm aware of the unused static method. I left it in there for the future, if the states of `blockDreamDashes` and `blockDreamTunnelDashes` would need to be known at the same time.  
   The method is there to avoid having to foreach twice.
2. `CollideAll<T>(Vector2 at)` calls the same method as `CollideAll<T>(Vector2 at, List<Entity> into)`. The first overload just instantiates a new list, while the second one just uses the list in the second parameter.
   As the code comment mentions, the list is kept around solely to prevent allocating a brand new list every time `CollideAll<T>` is called. The `into` parameter and return value are the same list, and are always reference equal.

PR #243, opened 2025-05-11
Add option to define death/respawn animation IDs to Glow Controller

aonkeeper4:
Adds the option to define death/respawn animation IDs for entities in Glow Controllers. When the targeted entities' Sprites play an animation in one of these lists, the light/bloom of that entity will fade out/in over the duration of the animation. This behaviour was previously special-cased for jellyfish.


catapillie (approved):
Thank you

PR #245, opened 2025-05-17
Make custom dream block particles more robust.

CPCTC:
Fulfills [this](https://discord.com/channels/835962669453410354/1373210219869634630/1373210219869634630) request.

- Adds a way to optionally customize the color of dream block particles
- Makes the default color of said particles respect the existing undocumented behavior when `Refill Count` is `-2`
- Document that setting `Refill Count` to `-2` makes the dream block conserve Madeline's dashes.


CPCTC:
Actually, I want to make a minor change to the Loenn plugin, so I'll re-open this as soon as I get around to that :gladeline:

PR #246, opened 2025-05-29
Fix Dream Sprites not being untracked properly when parent entity's depth is changed

aonkeeper4:
Fixes Dream Sprites not being assigned to the right renderer when their parent entity's depth is changed. This caused "ghost" Dream Jellyfish and Dream Theo Crystals to appear when the player died while holding one after being launched from a fish/bumper.

### DIFF
--- a/everest.yaml
+++ b/everest.yaml
@@ -1,5 +1,5 @@
 - Name: CommunalHelper
-  Version: 0.0.0-dev
+  Version: 1.23.0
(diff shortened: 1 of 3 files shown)


catapillie (approved):

catapillie:
Thank you for the fix, looks good. Not sure about the performance of hooking into a general setter like `Depth.set`, but I suppose it's negligible given entities don't change depth often (I think).

PR #247, opened 2025-05-30
Flag functionaliy for Manual Cassette controller

Etpio:
Currently, manual cassette controller (Communal Helper) and cassette flag controller (Crystalline Helper) are incompatible. After raising this issue to Vitellary, we came to the conclusion it would be best to add flag functionality to manual cassette controller.

The PR basically merges the two controllers together. There may be a better way to get the flag names field in loenn, the implementation I used is identical to the one in Crystalline however (and preserves the same names too)

Also the lua file for manual cassette controller referred to it as "jump fix controller" for some reason, so I fixed it on the way.

### DIFF
--- a/src/Entities/CassetteBlocks/ManualCassetteController.cs
+++ b/src/Entities/CassetteBlocks/ManualCassetteController.cs
@@ -14,20 +14,14 @@ public class ManualCassetteController : AbstractInputController
     private int roomBeats;
     private int currentIndex;
 
-    private string blueFlag;
-    private string pinkFlag;
-    private string yellowFlag;
-    private string greenFlag;
+    private const string blueFlag = "CH_cas_blue";
+    private const string pinkFlag = "CH_cas_rose";
+    private const string yellowFlag = "CH_cas_brightsun";
+    private const string greenFlag = "CH_cas_malachite";
 
     public ManualCassetteController(EntityData data)
     {
         startIndex = data.Int("startIndex", 0);
-              
-        // do not change those flags!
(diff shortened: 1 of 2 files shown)


src/Entities/CassetteBlocks/ManualCassetteController.cs
```diff
@@ -13,13 +14,29 @@ public class ManualCassetteController : AbstractInputController
     private int roomBeats;
     private int currentIndex;
 
+    private string blueFlag;
```
catapillie: These can be declared as constants strings, instead of private fields initialized in the constructor.
```suggestion
    private const string blueFlag = "CH_cas_blue";
```
Then you don't have to add those lines in the constructor (26 to 30).

src/Entities/CassetteBlocks/ManualCassetteController.cs
```diff
@@ -59,11 +76,45 @@ public void Tick()
     {
         currentIndex++;
         currentIndex %= roomBeats;
+        SetFlag(currentIndex);
         SetActiveIndex(currentIndex);
         Audio.Play("event:/game/general/cassette_block_switch_" + ((currentIndex % 2) + 1));
         Input.Rumble(RumbleStrength.Medium, RumbleLength.Short);
     }
 
+    public void SetFlag (int index)
```
catapillie: This function can be shorter, but eh that's fine I'll tweak it myself.

catapillie (approved):
Thanks! I'll just tweak the code slightly to make it shorter and sweeter.

PR #248, opened 2025-06-03
Add Retract Speed Modifier to Chained Kevins

karl2883:
Add retract speed modifier to chained kevins.
Implementation creates two new hooks, which isn't ideal, but there didn't seem to be a cleaner way to do it and the hooked methods (MoveTowardsX/MoveTowardsY) aren't actually used that often (MoveTowardsX only for Kevins, in fact).
Per request for a collab.

### DIFF
--- a/Loenn/entities/chains/chained_kevin.lua
+++ b/Loenn/entities/chains/chained_kevin.lua
@@ -33,6 +33,7 @@ for _, direction in pairs(directions) do
                 chillout = false,
                 chainLength = 64,
                 direction = direction,
+                retractSpeedModifier = 1.0,
                 chainOutline = true,
                 centeredChain = false,
                 chainTexture = "objects/CommunalHelper/chains/chain"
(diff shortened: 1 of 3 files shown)


catapillie:
Sorry, but I really don't like this hook into `MoveTowardsX` and `MoveTowardsY`. I guess the best way to do this would be to IL Hook into `Celeste.CrushBlock+<AttackSequence>d__41.MoveNext`, and insert the multiplication by the retract speed factor there.
That'd be around here:
```
// <speed>5__3 = Calc.Approach(<speed>5__3, 60f, 160f * Engine.DeltaTime);
	IL_0864: ldarg.0
	IL_0865: ldarg.0
	IL_0866: ldfld float32 Celeste.CrushBlock/'<AttackSequence>d__41'::'<speed>5__3'
	IL_086b: ldc.r4 60
// --- mul by `retractSpeedModifier` here
	IL_0870: ldc.r4 160
	IL_0875: call float32 Monocle.Engine::get_DeltaTime()
	IL_087a: mul
	IL_087b: call float32 Monocle.Calc::Approach(float32, float32, float32)
	IL_0880: stfld float32 Celeste.CrushBlock/'<AttackSequence>d__41'::'<speed>5__3'
```

I can implement it if you don't feel like doing the IL yourself. No worries!
Either way thanks for the idea and the partial implementation!

PR #249, opened 2025-06-06
Add Bounce On Collision, Revert On Death and Revert On Leave options to Configure Dream Tunnel Dash triggers

aonkeeper4:
Adds a new Bounce On Collision option to Configure Dream Tunnel Dash triggers.
Also reworks configure state triggers to work more like Extended Variant triggers, including new Revert On Death and Revert On Leave options.
Unfortunately, due to the implementation using generic classes (which aren't really supported by `[Tracked(true)]`), this PR bumps the Everest dependency to latest stable in order to use `[TrackedAs(typeof(UntrackedType))]`.


catapillie (approved):
Briefly tested here, seems to work well! Thanks for all of this, this looks great.

PR #250, opened 2025-06-29
Fix Charged Aero Block wallbounce leniency

earthwise01:
Adds an option to increase the boost leniency for wallbounces on Charged Aero Block to cover the full 5 pixels instead of only the 3 pixels a regular walljump has, which left a 2 pixel window where you could wallbounce but not receive the boost.

Ideally I would've made this retroactively affect existing aero blocks too, but from what I understand it's generally advised to avoid changing behaviour in existing maps as much as possible, so this is only enabled by default for newly placed aero blocks.

This also replaces some reflection used to modify the player's `varJumpSpeed` for stronger vertical boosts with directly accessing it using the publicizer.

### DIFF
--- a/src/Entities/Aero/AeroBlockCharged.cs
+++ b/src/Entities/Aero/AeroBlockCharged.cs
@@ -9,8 +9,6 @@ namespace Celeste.Mod.CommunalHelper.Entities;
 public class AeroBlockCharged : AeroBlockFlying
 {
     internal static bool SpirialisHelperLoaded = false;
-    private static readonly FieldInfo player_varJumpSpeed
-        = typeof(Player).GetField("varJumpSpeed", BindingFlags.NonPublic | BindingFlags.Instance);
 
     [Flags]
     private enum ButtonCombination : byte
@@ -381,7 +379,7 @@ public class AeroBlockCharged : AeroBlockFlying
             orig(self, particles, playSfx);
             if (self.OnGround() && block is not null && block.CheckTopButton())
                 block.Smash(self, Vector2.UnitY * -350);
-            player_varJumpSpeed.SetValue(self, self.Speed.Y);
+            self.varJumpSpeed = self.Speed.Y;
         }
         else
         {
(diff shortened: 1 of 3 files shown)


src/Entities/Aero/AeroBlockCharged.cs
```diff
@@ -147,13 +150,13 @@ public static Button TopButton(AeroBlockCharged entity, bool visible)
...
+        : this(data.NodesWithPosition(offset), data.Width, data.Height, data.Bool("loop"), data.HexColor("activeColor", defaultOnColor), data.HexColor("inactiveColor", defaultEndColor), data.Bool("hover", true), data.Attr("buttonSequence", DEFAULT_BUTTON_SEQUENCE), data.Bool("wallbounceLeniency", false))
     {
         // if the new attribute exists use that, but otherwise enable the bug if either the old attribute is set *or* spirialis helper is loaded
         SpirialisBug = data.Bool("SpirialisBugV2", SpirialisHelperLoaded || data.Bool("SpirialisBug", false));
     }
 
-    public AeroBlockCharged(Vector2[] positions, int width, int height, bool loop, Color activeColor, Color inactiveColor, bool hover = true, string buttonSequence = DEFAULT_BUTTON_SEQUENCE)
+    public AeroBlockCharged(Vector2[] positions, int width, int height, bool loop, Color activeColor, Color inactiveColor, bool hover = true, string buttonSequence = DEFAULT_BUTTON_SEQUENCE, bool wallbounceLeniency = false)
```
catapillie: Shouldn't it be `bool wallbounceLeniency = true`? I know it won't matter, but theoretically if someone were to instantiate this entity from within the code, wouldn't it make more sense for the option to be on by default?

catapillie (commented):
Thank you for the fix! Will merge later today

src/Entities/Aero/AeroBlockCharged.cs
earthwise01: oh yup I was just copying what I did in the other constructor to make existing placements not have the fix, but that would make more sense I suppose

catapillie (approved):

PR #251, opened 2025-06-29
Fix incorrect dream tunnel dash hook logic

aonkeeper4:
Fixes incorrect dream tunnel dash hook logic, for example in `Spring.OnCollide` that would cause the player to get stuck in the solid they were tunnelling through when hitting a spring.


catapillie (approved):
👍

PR #253, opened 2025-07-04
Fix can climb up wall from the bottom of the sideways TimedTriggeredSpikes

AAA1459:
https://github.com/user-attachments/assets/63514320-0437-47c7-b2d1-44608fe1a106

### DIFF
--- a/Loenn/entities/misc/timed_trigger_spikes.lua
+++ b/Loenn/entities/misc/timed_trigger_spikes.lua
@@ -17,6 +17,7 @@ for _, dir in ipairs(directions) do
         placement.data.Rainbow = false
         placement.data.Reusable = false
         placement.data.ReusableTimer = 0.0
+        placement.data.legacyBehavior = true
     end
 
     table.insert(timedTriggerSpikes, spikes)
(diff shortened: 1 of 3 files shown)


catapillie:
I'm unable to reproduce, could you give more info on what exactly is going on?

aonkeeper4:
I was able to reproduce - vanilla trigger spikes block the player from climbing up onto a ledge if they would climbhop into the trigger spikes. This only happens on dust trigger spikes though, not any of the other variants, but I had a look and it just seems like the option to enable the behaviour parity isn't in Loenn yet. Timed Trigger Spikes currently don't have this check.

aonkeeper4 (changes requested):
This is good, but I feel like the fixed behaviour should be under a toggle that defaults to false on old placements, as it is possible (though unlikely) that this change could break existing maps.

AAA1459:
> This is good, but I feel like the fixed behaviour should be under a toggle that defaults to false on old placements, as it is possible (though unlikely) that this change could break existing maps.

Yeah, it's been 8 months and I've almost forgotten about here.

AAA1459:
btw everest side: https://github.com/EverestAPI/Everest/pull/921

catapillie:
I would've merged it immediately because it's clearly a bug, but because Everest introduces a legacy option, I think we should also do it (to keep "parity")

AAA1459:
> I would've merged it immediately because it's clearly a bug, but because Everest introduces a legacy option, I think we should also do it (to keep "parity")

done

aonkeeper4:
Looks good, thanks!

PR #254, opened 2025-07-23
Add cassette functionality to Charged Aero Blocks

aonkeeper4:
Adds options to Charged Aero Blocks to allow syncing the activation of their buttons to the activation of a specified Cassette Block index, alongside refactoring their hooks to look less messy.


catapillie (approved):
Awesome idea and thank you for the refactor. This looks really cool, I like what you did with the visuals

PR #255, opened 2025-08-28
Chain camera culling

JaThePlayer:
Implements camera culling for chains.
Also:
- Removes linq from Update, instead opting to allocate `oldNodePositions` upfront in the constructor, then reusing the same array each frame instead of allocating a new array each frame.
- Adds new `updateOffscreen` option (defaults to true). If disabled, camera culling also disables `Update` logic. Because this could prevent chains from breaking if they go offscreen, this has to be manually changed by the mapper if they decide this is not an issue in their usecase.

### DIFF
--- a/src/Entities/Chains/Chain.cs
+++ b/src/Entities/Chains/Chain.cs
@@ -1,5 +1,5 @@
-﻿using FMOD.Studio;
-using System.Linq;
+﻿using Celeste.Mod.Helpers;
+using FMOD.Studio;
 
 namespace Celeste.Mod.CommunalHelper.Entities;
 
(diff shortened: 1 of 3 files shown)


src/Entities/Chains/Chain.cs
```diff
@@ -262,9 +272,28 @@ private void ShakeImpulse()
...
+        Vector2 end = nodes[^1].Position;
+
+        float x = Math.Min(begin.X, Math.Min(control.X, end.X));
+        float right = Math.Max(begin.X, Math.Max(control.X, end.X));
+        float y = Math.Min(begin.Y, Math.Min(control.Y, end.Y));
+        float bottom = Math.Max(begin.Y, Math.Max(control.Y, end.Y));
+
+        return CullHelper.IsRectangleVisible(x, y, right - x, bottom - y, lenience: segment.Width + 2);
+    }
+
```
catapillie, line 290: I don't think that this is mathematically correct. The starting, ending and middle chain nodes don't necessarily define the chain's bounds, because it can take any wild shape caused by moving solids, or player interaction, or whatever. I'm convinced enough that this is a good approximation (given the leniency parameter as well), so it's fine, just needed to point it out

catapillie (commented):
Thank you, that looks good

PR #258, opened 2025-09-07
Add reskin support for custom cassette entities

aonkeeper4:
does what it says on the tin. also allows changing the alpha of cassette block sides


catapillie (approved):
I think it'd be better if each cassette block type had its own specific instructions for reskinning in the tooltips. Otherwise, everything else is ok

PR #259, opened 2025-09-22
Allow adding cloudscapes as foreground stylegrounds

earthwise01:
Also sets the rasterizer state to `RasterizerStates.CullNone` before rendering to prevent it from completely exploding sometimes (this would happen if a cloudscape was the first fg styleground since the Bloom.Apply call immediately before fg styleground rendering leaves it set to `RasterizerStates.CullCounterClockwise`)


catapillie (approved):
Thank you!

PR #263, opened 2025-11-24
Fix old configure state triggers having "revert on death" behaviour by default

aonkeeper4:
Due to an oversight, configure state triggers placed before 1.24.0 have the "revert on death" behaviour enabled by default, breaking maps like An Artificer's Ascent from SSC3. This PR fixes the bug, and also makes the revert on death/leave behaviour function as expected, as previously you would be able to trick it into not doing anything by entering the trigger twice. SpeedrunTool interop has also been updated to account for all static fields in player state classes.
Ideally this should be merged alongside #257 and #262, just to have all the state-related fixes in the same update.


LozenChen:
in srt interop, could possibly change `"hasSeekerDash"` to `nameof(SeekerDash.hasSeekerDash)` so it won't break after rename?

plz remind me when this PR gets merged so i can remove related codes in srt

PR #264, opened 2026-01-03
Add option to disable the pulsing of Move Block Groups

aonkeeper4:
This PR adds an option to Move Block Groups to disable the pulsing of the arrows on Move Blocks in the group.


catapillie (approved):
perfect, thank you!

PR #265, opened 2026-01-03
Fix Shapeshifter rumble sound persisting

aonkeeper4:
Before, the `quakeSfx` for Shapeshifters was not properly stopped when the Shapeshifter was removed from the scene. This PR ensures that this sound is always properly stopped.


catapillie (approved):
LGTM

PR #266, opened 2026-01-03
Add Disable Auto Camera Offset Controller

aonkeeper4:
This PR adds a controller that disables the automatic camera offset that is applied when the player is in some states, such as `StRedDash` (red boosters, also Communal Helper's own custom boosters) and `StStarFly` (feather state), optionally being enabled/disabled by a flag.


aonkeeper4:
Implemented in aon helper 0.13.0.

PR #268, opened 2026-03-05
Fix lags when loading any room with a lot of decals (and entities)

AAA1459:
https://discord.com/channels/403698615446536203/1478910318498283621


aonkeeper4:
Discussed in linked thread.

PR #269, opened 2026-03-07
Refactor/fix dash collision in panels

karl2883:
This moves all dash collision related logic of the AbstractPanel class into the new AbstractDashCollisionPanel class, which the panels who need that logic (DashCollisionPanel, DreamTunnelEntry) inherit instead.

Rationale: In the previous version, the dash collision logic of AbstractPanel ran for all panels, even those who do not need it. The logic seems like it should not affect anything, but it does: When the OnDashCollision hook gets applied through the call to DelegateHelper, the original method (OnDashCollision) gets replaced with a default method returning NormalCollision if it was previously null. Having a NormalCollision or having no OnDashCollision method actually behaves differently - mainly in the case of red bubbles, which do not exit if there is a NormalCollision and do exit if there is no collision. So, the OnDashCollision method should only be hooked/overridden if the panel does in fact want to modify OnDashCollision. In that case, behaviour change like for the red bubbles is to be expected.

### DIFF
--- a/src/DashStates/DreamTunnelDash/DreamTunnelEntry.cs
+++ b/src/DashStates/DreamTunnelDash/DreamTunnelEntry.cs
@@ -27,7 +27,7 @@ namespace Celeste.Mod.CommunalHelper.Entities;
 
 [CustomEntity("CommunalHelper/DreamTunnelEntry = LoadDreamTunnelEntry")]
 [Tracked]
-public class DreamTunnelEntry : AbstractPanel
+public class DreamTunnelEntry : AbstractDashCollisionPanel
 {
     private struct DreamParticle
     {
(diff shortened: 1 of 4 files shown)


karl2883:
If this gets merged, please credit @lyraohmi on Discord as well, as they did the painful work of tracking down where the bug was :)

aonkeeper4 (commented):
I agree with splitting the dash collision overriding logic into its own subclass, but shouldn't this also fix the offending `NormalCollision` default in `DelegateHelper` since that's the actual root of the problem? Changing it to `NormalOverride` should do the trick.

karl2883:
The issue is that the problem in DelegateHelper is kind of unfixable as far as I can tell. Changing it to `NormalOverride` will not fully do the trick, as it does not affect vertical collisions. So I think that the best course of action is limiting this behaviour to entities for which dash collision is important.

aonkeeper4:
Interesting, I suppose that this can be merged then. Thanks for the fix.

PR #270, opened 2026-03-14
Add Elytra Note Ring customization

SunsetQuasar:
Adds custom events, note volume, ring color, and support for passing audio parameters.

Plugin untested in Lönn \**wink wink nudge nudge*\*, but probably works.

Unless I fucked up really bad, all previous placements should remain intact.

https://github.com/user-attachments/assets/8aaa2baa-430e-4983-b636-fb8a7c1f70c6


SunsetQuasar:
Actually: should the event be stored in a SoundSource Component? Currently if the event played is a continuous one, you get the evil perpetual stackable event issue

aonkeeper4 (approved):
Looks good, thanks!

PR #271, opened 2026-03-15
Fix Chained Falling Block static mover shake time + add option to render the chain behind

earthwise01:
Adds an option to fix Chained Falling Blocks not shaking for their full shake time when activated by attached static movers (like other falling blocks do), and an option to make just the chain render farther behind in the scene.


aonkeeper4 (approved):
This all looks great, thanks!

PR #272, opened 2026-03-15
Add `Shape3D` and `Shapes` ModInterop

SunsetQuasar:
Self-explanatory. Exposes methods from `Components.Shape3D` and `Utils.Shapes` as `Component`s/`object`s respectively so you won't need an assembly reference.


aonkeeper4 (approved):
This all looks good. I'll port over the stuff for loading custom models in a later commit.

PR #275, opened 2026-04-07
Add support for entity SIDs to Glow Controllers and Show Hitbox Triggers

earthwise01:
This PR adds support for using entity SIDs instead of class names to Glow Controllers and Show Hitbox Triggers, and adds dropdowns with all SIDs in the current map to their Lönn plugins.

Also makes Glow Controllers use the tracker instead of searching through the entire entity list (should be sufficient since entities are added to the tracker earlier in `UpdateLists` than where we're hooking) and process entities inside the normal `toAwake` loop.


aonkeeper4 (approved):
Looks great, thanks for the improvement!

PR #279, opened 2026-05-19
Zoom compatibility for Lightning Controller

ozmarie:
This makes Lightning Controller render its flash and bolts relative to the camera rather than an absolute 320x180, which should make it compatible with things like Extended Camera Dynamics.


aonkeeper4 (changes requested):
the evil zip ... code looks good though

ozmarie:
evil zip is dead

aonkeeper4:
yay awesomesauce :tada: tested and works

PR #286, opened 2026-09-13
use color list fields where appropriate

earthwise01:
This PR makes the Lönn plugins for Cloudscapes, Beta Cubes, and Cloudscape Color Transition + Lightning Configuration Triggers use color list fields for their options that accept comma separated lists of colors.


catapillie (approved):
Thank you, glad this is a Lönn feature now (probably been for a while, I just never noticed)

